PR #8, opened 2020-03-04
Initial Windows packer image build

slide:


Jenkinsfile
```diff
@@ -23,6 +23,7 @@ pipeline {
     stage('Validate') {
       steps {
         sh 'packer validate --var-file validate-vars.json ubuntu-18-agent.json'
+        sh 'packer validate --var-file validate-vars.json windows-2019-agent.json'
```
timja: Would this be a good use case for a matrix based pipeline? https://jenkins.io/blog/2019/11/22/welcome-to-the-matrix/

windows-2019-agent.json
```diff
@@ -0,0 +1,51 @@
...
+      "winrm_insecure": true,
+      "winrm_username": "packer",
+      "winrm_timeout": "20m",
+
+      "client_id": "{{user `client_id`}}",
+      "client_secret": "{{user `client_secret`}}",
+      "subscription_id": "{{user `subscription_id`}}",
+
+      "location": "{{user `location`}}",
+      "managed_image_name": "jenkins-agent-win2019-{{isotime \"20060102150405\"}}",
```
olblak: I know we are doing the same for the ubuntu template but wouldn't it be better to **not** use isotime in the name so we don't always have to update our jenkins configuration (ci, trusted,...). If for some reasons the image is broken, we have to fix it as soon as possible anyway. 
Or we could build two images, one latest and one with a specific tag.
Personnally I am more in favor for the latter

olblak: @timja any opinions?

timja: I would just try removing the isotime from it, and see how it goes, much nicer if we don't have to update it the other option is a shared image gallery, we could automatically add it to the gallery here and update the latest tag, but that would still allow downgrading if required

timja:
```suggestion
      "managed_image_name": "jenkins-agent-win2019",
```

timja:
@bitwiseman Am I doing something wrong here?
image

The only way I can tell which axe of the matrix that is running is by inspecting the html and I can see its there but blueocean is cutting it off, I also assumed I would be able to interpolate the stage name with the axes but it doesn't seem to work.

<details>

<summary>Pipeline</summary>

```groovy

if (env.CHANGE_ID) {
  properties([
    buildDiscarder(logRotator(numToKeepStr: '10')),
  ])
} else {
  properties([
    buildDiscarder(logRotator(numToKeepStr: '96')),
    pipelineTriggers([[$class: "SCMTrigger", scmpoll_spec: "H/10 * * * *"]]),
  ])
}

pipeline {
  agent none

  stages {
    stage('ValidateAndBuild') {
      matrix {
        agent {
          docker {
            args '--entrypoint=""'
            image 'hashicorp/packer:1.5.4'
            label 'docker&&linux'
          }
        }
        axes {
          axis {
            name 'AGENT'
            values 'ubuntu-18', 'windows-2019'
          }
        }
        stages {
          stage('Validate ${AGENT}') {
            steps {
              sh 'packer validate --var-file validate-vars.json ${AGENT}-agent.json'
              sh 'packer validate --var-file validate-vars.json ${AGENT}-agent.json'
            }
          }

          stage('Build ${AGENT}') {
            environment {
              AZURE_SUBSCRIPTION_ID = credentials('packer-azure-subscription-id')
              AZURE_CLIENT_ID = credentials('packer-azure-client-id')
              AZURE_CLIENT_SECRET = credentials('packer-azure-client-secret')
            }

            when {
              branch 'master'
            }

            steps {
              sh """
                packer build \
                  --var location="East US 2" \
                  --var resource_group_name="prod-packer-images" \
                  --var subscription_id="$AZURE_SUBSCRIPTION_ID" \
                  --var client_id="$AZURE_CLIENT_ID" \
                  --var client_secret="$AZURE_CLIENT_SECRET" \
                  ${AGENT}-agent.json
              """
            }
          }
        }
      }
    }
  }
}


```

</details>

bitwiseman:
@timja 
Stage names must be literal strings.  Making them not have to be is likely to be very difficult. That is why
`stage('Validate ${AGENT}'` is correct syntax but also doesn't interpolate.  

The parallel branch name being cut off  in Blue Ocean is annoying, but is up to others to fix.

PR #14, opened 2020-03-11
Add git-lfs to Windows image and a Windows docker image

slide:
Added git-lfs to the Windows image and added a Windows docker image.


MarkEWaite:
I confirmed the image download locations for git and for git LFS are correct. I also learned about MinGIT for the first time.  I use Portable Git (with the exe installer).  Will be good to have MinGIT in CI so that we're testing both MinGIT (CI) and Portable Git (my env)

PR #15, opened 2020-03-11
Add initial AWS images (ubuntu 18.04, windows 2019, windows 2019 + docker)

slide:


ubuntu-18-agent-aws.json
```diff
@@ -0,0 +1,62 @@
...
+      "tags": {
+        "timestamp": "{{isotime \"20060102150405\"}}",
+        "imagetype": "jenkins-agent-ubuntu18.04"
+      }
+    }
+  ],
+  "provisioners": [{
+    "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
+    "inline": [
+      "useradd -m -o -u 1000 jenkins",
```
olblak: Isn't already used by the ubuntu user?

slide: Yes, the `-o` means that the UID doesn't need to be unique.

olblak:
The Jenkinsfile also needs to be refactored in order to be able to push images to AWS, I propose to create two directories aws and azure and move packer files in their respective directories

timja:
> The Jenkinsfile also needs to be refactored in order to be able to push images to AWS, I propose to create two directories aws and azure and move packer files in their respective directories

makes sense

ubuntu-18-agent-aws.json
```diff
@@ -0,0 +1,62 @@
...
+
+      "tags": {
+        "timestamp": "{{isotime \"20060102150405\"}}",
+        "imagetype": "jenkins-agent-ubuntu18.04"
+      }
+    }
+  ],
+  "provisioners": [{
+    "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
+    "inline": [
```
timja: is is worth extracting this out to a script? I don't see any changes between aws and azure here?

olblak: It would a nice improvement but not mandatory IMHO.

timja: should be straightforward since windows is done like this

slide:
> The Jenkinsfile also needs to be refactored in order to be able to push images to AWS, I propose to create two directories aws and azure and move packer files in their respective directories

I started on updating the Jenkinsfile, but didn't want to push anything without the credentials in there. I can change it to use directories for sure.

ubuntu-18-agent-aws.json
slide: Should be pretty easy, I can pull it out as part of this.

timja:
> > The Jenkinsfile also needs to be refactored in order to be able to push images to AWS, I propose to create two directories aws and azure and move packer files in their respective directories
> 
> I started on updating the Jenkinsfile, but didn't want to push anything without the credentials in there. I can change it to use directories for sure.

creds should be there, packer-aws

olblak:
> packer

I replace the aws credentials by `packer-aws-secret-access-key` and `packer-aws-secret-access-key`

PR #28, opened 2020-03-26
Trigger Jenkinsfile every Sunday on master branch

olblak:
We don't need to generate new images every 10 minutes


olblak:
Good point

PR #29, opened 2020-04-11
Merge the Windows images

slide:
Since we can use WindowsCore as base for both, we can have just one image for Windows that can be used for maven builds and docker builds.


timja:
Builds failing

PR #31, opened 2020-05-28
Add windows-10 agent for Azure

slide:
This could be used for testing for the Windows Support Policy.

### DIFF
--- a/azure/windows-10-agent.amd64.json
+++ b/azure/windows-10-agent.amd64.json
@@ -1,0 +1,57 @@
+{
+  "variables": {
+    "maven_version": "3.6.3",
+    "git_version": "2.25.1",
+    "jdk11_version": "11.0.6+10",
+    "jdk8_version": "8u242-b08",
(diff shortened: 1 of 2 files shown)


azure/windows-10-agent.amd64.json
```diff
@@ -0,0 +1,57 @@
+{
+  "variables": {
+    "maven_version": "3.6.3",
+    "git_version": "2.25.1",
+    "jdk11_version": "11.0.6+10",
+    "jdk8_version": "8u242-b08",
```
oleg-nenashev, line 6: Update to latest?

slide: I was keeping it the same as the other Windows packer images for now, we can update all of the images to latest in another PR. Is that ok with you?

olblak: It sounds good to me, and I am also in favor to use the same version everywhere

PR #39, opened 2020-10-13
Only run most recent build of a given branch

jglick:


markyjackson-taulia (approved):

garethjevans:
did not know you could do that!

jglick:
Really there ought to be a nicer way of doing it, like say a job property (`properties`) and/or a GUI/CasC-configured branch property. This weird syntax is the current workaround. See Jira issues.

PR #55, opened 2021-04-21
Chore: add a daily cloud cleanup task [INFRA-2958]

dduportal:
This PR introduces the following changes:

- Add a cleanup task, executed at the end of the build. This task is executed in "dry run" all the time, and "for real" on the principal branch. There are 2 types of cleaned resources for now: instances and security groups.
- Fixes Pipeline to support the principal branch name change to `main`
- Switch Pipeline execution to Kubernetes Agent instead of Docker container in EC2 VMs (faster and cheaper)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,16 +1,8 @@
-def buildNumber = BUILD_NUMBER as int; if (buildNumber > 1) milestone(buildNumber - 1); milestone(buildNumber) // JENKINS-43353 / JENKINS-58625
 
-if (env.CHANGE_ID) {
-  properties([
-    buildDiscarder(logRotator(numToKeepStr: '10')),
-  ])
-} else {
-  properties([
-    buildDiscarder(logRotator(numToKeepStr: '96')),
- // Disable for now until we fix https://ci.jenkins.io/job/Infra/job/packer-images/job/master/148/console
- //   pipelineTriggers([[$class: "SCMTrigger", scmpoll_spec: "H/10 * * * *"]]),
-  ])
-}
+properties([
+  buildDiscarder(logRotator(numToKeepStr: '10')),
+  pipelineTriggers([cron('@daily')]),
+])
 
 pipeline {
   agent {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -1,16 +1,8 @@
...
-} else {
-  properties([
-    buildDiscarder(logRotator(numToKeepStr: '96')),
- // Disable for now until we fix https://ci.jenkins.io/job/Infra/job/packer-images/job/master/148/console
- //   pipelineTriggers([[$class: "SCMTrigger", scmpoll_spec: "H/10 * * * *"]]),
-  ])
-}
+properties([
+  buildDiscarder(logRotator(numToKeepStr: '10')),
+  pipelineTriggers([cron('@daily')]),
```
timja, line 4: you want PRs to trigger this?

Jenkinsfile
```diff
@@ -1,16 +1,8 @@
-def buildNumber = BUILD_NUMBER as int; if (buildNumber > 1) milestone(buildNumber - 1); milestone(buildNumber) // JENKINS-43353 / JENKINS-58625
```
timja, line 1: why are you removing this?

Jenkinsfile
dduportal: yes, the intent is to execute this pipeline:

- On PRs (when a webhook is received from GitHub)
- Once a day

I've used the Pipeline syntax snippet generator to get this line: did I miss something?

Jenkinsfile
dduportal: Because I have no idea what is it for, and it's documented nowhere?

timja: it's not, it's a @jglick jenkins hack to do with parallel builds and aborting old ones when new ones come in or something along those lines

dduportal: Oh I see: it explains why my builds where always killed 😅 
Then removing this line is a feature then: it allows to not waste builds when doing extensive tests (given the build time here: I would prefer not killing previous builds).

WDYT?

dduportal: (I would also add that by switching to Kubernetes pods instead of VMs, we are not wasting a lot of resource with multiple builds now, WDYT?)

Jenkinsfile
jglick: The `cron` trigger presumably makes sense only when `branch == 'main'`.

Jenkinsfile
jglick: Up to you. When builds are long and resources are limited, aborting outdated builds is generally best.

Jenkinsfile
dduportal: @jglick good point ❤️

Jenkinsfile
dduportal: Good catch @timja @jglick ❤️ 

With the explanation from both of you, I confirm that the PR is kept with this deletion.

timja: It was introduced in https://github.com/jenkins-infra/packer-images/pull/39

I think because of lots of packer builds holding a lot of executors in ci.jenkins.io...

dduportal: @timja yup, thanks! I also ended on the same PR (39) from my research to understand a bit better. But as the current PR also introduces kubernetes agent, the executor limitation is not there anymore

timja (approved):

PR #56, opened 2021-04-22
chore: factorize templates per clouds

dduportal:
This PR introduces the following changes:

- Contains the configuration & code spread by factorizing the packer templates per cloud provider. This is a direct usage of the HCL switch, and ensure that there are less and less duplication of code and configuration.
  - Note that there is still a spread between cloud providers (the software versions for instance). This duplication between cloud should be a subsequent PR.
- Simplify the build script:
  - Keeping a shell script instead of using a Makefile, to allow using directly the Docker image `hashicorp/packer`which does not have `make` installed. 
  - Compromise to make the Jenkinsfile less dependent on the commands, and providing reproducibility for a local build


README.md
```diff
@@ -2,23 +2,57 @@
...
 A repository containing a number of packer build configurations for AWS & Azure.
 
-## Invoking the Azure builds locally
+## Requirements
 
-Create a new resource group
+* A shell (sh, bash, ash, zsh) prompt
+* Packer (check template's version constraints to know which version to use)
+* Define the environment variables `PKR_VAR_cloud`, `PKR_VAR_agent` and `PKR_VAR_architecture` to the target you want to build
+* Configure the cloud (defined on the variable `PKR_VAR_cloud`) credentials (e.g. APi key, or Cloud's CLI access sur as `az` or `aws` commands))
```
timja, line 10:
```suggestion
* Configure the cloud (defined on the variable `PKR_VAR_cloud`) credentials (e.g. APi key, or Cloud's CLI access such as `az` or `aws` commands))
```

timja (approved):

PR #58, opened 2021-05-04
Add more regions to the builds

dduportal:
Signed-off-by: Damien Duportal <damien.duportal@gmail.com>


Jenkinsfile
```diff
@@ -29,9 +29,35 @@ pipeline {
             name 'PKR_VAR_cloud'
             values 'aws', 'azure'
           }
+          axis {
+            name 'PKR_VAR_location'
+            values 'us-east-2', 'East US 2', 'East US'
```
timja, line 34: this is a bit confusing to me with all the excludes =/

dduportal: yes, I agree :'( I've added comment for each as a "best effort". 

We are limited by the current syntax. Evolutions could be:

* Split in 2 matrix (1 for AWS, 1 for Azure) and Build sequentially. But slow :'(
* Split in 2 matrix (1 for AWS, 1 for Azure) when declarative syntax would allow parallel stages with a matrix within
* Switch to scripted syntax, but I'm not willing to spend time on this unless really mandatory. My brain is not wired for scripted pipelines.

Maybe others solutions?

olblak: Would you have a better solution? I don't see an easy solution to support different Cloud,Region,Platform

MarkEWaite (approved):

Jenkinsfile
timja: adding a comment here as well? appropriate location will be picked per cloud provider?

but no i don't have a better solution

dduportal: @timja LGTM, let me add a comment before merging

dduportal, line 34:
```suggestion
            // Aggregated list of all regions for all clouds. Look at the exclusions to know which location is for which cloud.
            values 'us-east-2', 'East US 2', 'East US'
```

PR #62, opened 2021-05-11
WIP: multi-regions + fix windows

dduportal:


azure/agent.pkr.hcl
```diff
@@ -42,46 +39,58 @@ variable "openssh_version" {
...
+  default = ""
 }
 variable "subscription_id" {
-  type = string
+  default = ""
+  type    = string
+}
+variable "image_version" {
+  type    = string
+  default = "0.0.2"
```
timja, line 55: do we have to bump this if we update the image? what's the workflow?

dduportal: - For this PR: no need to bump: rebuilding the image will override the existing image in version `0.0.2`, same a current behavior. It's only required to allow using collection.
- Plan: upcoming PR to use `jx-next-version` to determine and bump this value (and add automatic tagging)

WDYT?

timja: both work thanks

PR #67, opened 2021-05-25
feat: define a semantic version from git tags

dduportal:
This PR introduces semantic versioning of the images:

* Azure Shared Gallery is now fed the semantic version
* The images (on both Amazon and Azure) all have a tag `version` set to the version's value, to help tracking and cleaning up

The `jx-release-version` command line is used: please note that there is a custom pipeline shell step to gather the requirements. This steps might be replaced later by a custom image.

* `curl` and `git` are installed
* `git` is used to fetch tags (as the job configuration, right now, does not)
* `curl` is used to retrieve `jx-release-version` latest command

The scope of this PR is NOT about automatic tagging/release: this is the work of the next one :)

### DIFF
--- a/jenkins-agent.pkr.hcl
+++ b/jenkins-agent.pkr.hcl
@@ -56,9 +56,9 @@ variable "azure_subscription_id" {
   default = ""
   type    = string
 }
-variable "azure_image_version" {
+variable "image_version" {
   type    = string
-  default = "0.0.2"
+  default = "0.0.1" # Default is a valid version to not fail azure validation
 }
 variable "image_type" {
   type        = string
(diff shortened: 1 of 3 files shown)


jenkins-agent.pkr.hcl
```diff
@@ -56,9 +56,9 @@ variable "azure_subscription_id" {
   default = ""
   type    = string
 }
-variable "azure_image_version" {
+variable "image_version" {
   type    = string
-  default = "0.0.2"
+  default = "0.0.1" # Default is a valid version to not fail azure validation
```
olblak, line 61: This seems weird to me to replace 0.0.2 by 0.0.1

olblak (approved):

jenkins-agent.pkr.hcl
dduportal: Good point. The reason is that we started with the version "raw" in the code, and now the paradigm changed.
WDYT about setting the initial default value as `0.1.0` instead?

PR #72, opened 2021-06-24
Fix git-lfs installation for ARM64

dduportal:


dduportal:
Full test with build done here: https://ci.jenkins.io/blue/organizations/jenkins/Infra%2Fpacker-images/detail/PR-72/3/pipeline/108 (history rewritten since then)

PR #75, opened 2021-07-22
Setup QEMU

timja:


scripts/ubuntu-20-provision.sh
```diff
@@ -61,6 +61,15 @@ apt-get install -y --no-install-recommends \
...
   parallel
 
+# setup qemu
+apt-get install -y --no-install-recommends \
+  qemu \
+  binfmt-support \
+  qemu-user-static
+
+docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+docker run --rm -t arm64v8/ubuntu uname -m # test that qemu is working
```
timja, line 71:
```suggestion
```

scripts/ubuntu-20-provision.sh
```diff
@@ -61,6 +61,14 @@ apt-get install -y --no-install-recommends \
   jq \
   parallel
 
+# setup qemu
+apt-get install -y --no-install-recommends \
+  qemu \
+  binfmt-support \
+  qemu-user-static
+
+docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
```
dduportal, line 70: @timja shouldn't we run this command in the pipelines, to have Qemu loaded only when we need it?

timja: It just configures qemu, what would the reason of moving it be?

dduportal: I'm not comletely sure that the setting is persisted across reboot: let's check as it does not create any issue (I was overcomplicating)

dduportal (approved):
Let's go go go !

PR #81, opened 2021-08-29
chore: improve ubuntu provisioning script for sustainability

dduportal:
- Automatically define Ubuntu's code name when needed
- Keep using `apt-get` instead of `apt` (because unstable CLI API)
- Add CLI requirements check as first step to ensure early failure

This PR is a set of non functionnal improvements made along the recent changes here. It also sets the grounds for futur "Docker" images.


dduportal:
I'm merging this as it is non functional and minor

PR #85, opened 2021-09-01
chore: use infra.ci.jenkins.io instead of ci.jenkins.io

dduportal:
Why using infra.ci?

* Private instance means safer credentials as we interact with cloud (and generate cloud costs)
* Finer control of the job properties and traits
* Separation of concerns: ci.jenkins.io is meant for another persona than Jenkins infra team members

PR #88, opened 2021-09-15
Add a CONTRIBUTING.md documentation file

dduportal:
As per https://github.com/jenkins-infra/packer-images/pull/85#discussion_r709238282, this PR moves the documentation block related to contribution into a single file ./CONTRIBUTING.md to be visible on GitHub PRs.

Depends on #85


CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+# Contributing
+
+If you have:
+
+* Any question about this repository, please ask us on the IRC channel at #jenkins-infra or on <https://community.jenkins.io>
+
+* A problem, a feature request or an unwanted behavior with this code or its artifacts, please open an issue on the GitHub issue tracker of this repository.
+
+* A contribution to make, please fork the GitHub repository to your own account, create a named branch and open a Pull Request with your proposed change/fix
+  * Commit messages should follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) messages
```
timja, line 10: why? are we using them in some way?

is there a reason we aren't using release drafter like most jenkinsci repositories?

dduportal: * The idea behind using conventional commits is to allow automatic tag definition (based on the commit messages, we can automate the version increment without even thinking about it). This is not enabled/used here, but could be in the future.
* There isn't any reason for not using release-drafter: we can totally start using it to ensure a clean release notes,it's not mutually exclusive with the conventionnal commit \o/

timja: release drafter (I think) / a github workflow can use convention commits to automatically label

> The idea behind using conventional commits is to allow automatic tag definition (based on the commit messages, we can automate the version increment without even thinking about it). This is not enabled/used here, but could be in the future.

I don't think we should be complicating things that much for this repo.

99% of the time it will be just one change, we don't want to batch changes up as that will increase the risk.

Ideally everything is automatically released without even having to tag.
just using an increasing number like the jenkins build number or the repo commit count would probably work better imo

dduportal: Fair point. The only constraint is regarding the Azure Gallery which requires semantic versions (hence the versioning constraing put here).

But if we do not use conventional commit + enable release-drafter, that would make it simple:: is my understanding correct?

dduportal: > Ideally everything is automatically released without even having to tag.

I don't mind doing fully automatic build/rebuild but the downisde would be more noise on the PR opened on jenkins-infra for updating the agents: maybe we can "force" weekly rebuilds?

Was thinking to apply automatic releasing (like in docker-jenkins-* repos) as a weekly basis, if the 2-3 first releases works as expected: does it make sense?

timja: based on:
https://docs.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries#image-versions

it needs to be a version number with 3 parts.

but you can just keep bumping the 3rd number forever

> But if we do not use conventional commit + enable release-drafter, that would make it simple:: is my understanding correct?

I don't quite understand the question

dduportal:
Closing as per #85

PR #92, opened 2021-09-16
fix: Windows builds speed + sysprep

dduportal:
This PR introduces the following fixes to the windows builds and provisioning:

- Increase instance size to have enough I/O and network
- Add multiples instances of `windows-updates` as per https://github.com/rgl/packer-plugin-windows-update/issues/90#issuecomment-842276886 (tip from the author of the packer plugin to handle all windows updates edge cases)
  - implies using the provisioner's attribute `pause_before`
- Improve download performances on powershell script by disabling the progress bar (the gain is impressive: from 6m to 40s for downloading JDK11 for instance)
- Stop waiting for sys prep to reach a certain state: sometimes it's never reached until the VM is restarted, and there are no impacts in an environment such as this one (issue with the build stuck on loop with the message `IMAGE_STATE_COMPLETE` printed out)
- Putting back the timeout to 1 hour

The new build time shrank from ~39-40min to 29-30 min which is a win \o/

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -10,8 +10,8 @@ if (env.BRANCH_NAME == 'main') {
 pipeline {
   agent none
   options {
-    // Average build time is ~40 min. 2 hours timeout indicates that something is wrong and should fail
-    timeout(time: 2, unit: 'HOURS')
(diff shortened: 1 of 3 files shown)


timja (approved):
if it works

dduportal:
> if it works

Let's wait 30 more minutes :trollface:

PR #102, opened 2021-09-30
chore [updatecli] pin git-lfs to major version 2.x.y

dduportal:
This PR closes #99 by only checking for 2.x versions of git-lfs.


timja (approved):

MarkEWaite:
My initial tests of git LFS 3.0.1 confirm that the immediate failure of 3.0.0 is resolved.  I'll continue exploratory testing and report results in the infra IRC channel or here.

PR #103, opened 2021-09-30
Feat: Define a common paths for JDKs, only based upon major versions

dduportal:
This PR defines static installation paths for the JDK installation, which only depends on the java major version, to avoid the maintenance burden.

Ref. https://github.com/jenkins-infra/pipeline-library/issues/228#issuecomment-930219117.

The new paths are:

- Ubuntu Linux 20.04: 

```shell
/opt/jdk-8
/opt/jdk-11
/opt/jdk-17
```

- Windows 2019:

```powershell
C:\Tools\jdk-8
C:\Tools\jdk-11
C:\Tools\jdk-17
```

### DIFF
--- a/scripts/ubuntu-20-provision.sh
+++ b/scripts/ubuntu-20-provision.sh
@@ -10,7 +10,7 @@ echo "ARCHITECTURE=${ARCHITECTURE}"
 echo "COMPOSE_VERSION=${COMPOSE_VERSION}"
 echo "MAVEN_VERSION=${MAVEN_VERSION}"
 export DEBIAN_FRONTEND=noninteractive
-ubuntu_codename="$(cat /etc/os-release | grep UBUNTU_CODENAME | cut -d = -f 2)"
+ubuntu_codename="$(grep UBUNTU_CODENAME /etc/os-release | cut -d = -f 2)"
 
 ## Check for presence of requirements or fail fast
 for cli in add-apt-repository apt-get apt-cache awk curl grep groupadd head tar uname useradd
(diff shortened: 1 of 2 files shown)


timja (approved):

dduportal:
cc @jglick for information

scripts/ubuntu-20-provision.sh
```diff
@@ -110,8 +110,10 @@ rm -rf /tmp/git-lfs*
 apt-get install -y --no-install-recommends fontconfig
 
 ## OpenJDKs: Adoptium - https://adoptium.net/installation.html
+mkdir -p /opt/jdk-8 /opt/jdk-11 /opt/jdk-17
+
 # JDK8
-jdk8_short_version="$(echo "${JDK8_VERSION}" | sed 's/-//g')"
+jdk8_short_version="${JDK8_VERSION//-/}"
```
jglick, line 116: https://stackoverflow.com/a/13210909/12916 TIL, but is there a reason this looks different from the JDK 11 and 17 versions?

dduportal: Outside the non-coherent versioning schemes, I don't know the reason behind their choice.

The code is empirical, based on what we had on the former AdoptOpenJDK download links (and that are the same with Temurin).

You can see the 2 different patterns by looking at https://github.com/jenkins-infra/packer-images/blob/main/.auto.pkrvars.hcl#L3-L5 , and then by checking the download links on https://adoptium.net/ => JDK8 does not have any "separator character" at all, while JDK 11 and 17 have underscores in place of +.

PR #104, opened 2021-09-30
fix [azure windows-2019] use smaller base OS image to allow OS disk <100Gb

dduportal:
This PR uses the "smalldisk" version of Windows server Core 2019 with Container generation 2 (yes that is a loooong name).

This base OS features a system drive at 30 Gb instead of 127 Gb, leaving us a lot of room for builds (~75 Gb), but staying below the limit of 100 Gb disks.


timja (approved):
o nice, we should probably also look at upgrading to 2022 at some point

PR #106, opened 2021-10-04
Allow git LFS upgrade

MarkEWaite:
## Allow git LFS upgrade

Testing of git LFS 3.0.1 has been successful in all the cases that I have tested.  The blocking issue in git LFS 3.0.0 is resolved.

This reverts commit ff6e722173039e09a8f41243fb5d22656b34322a.

### DIFF
--- a/updatecli/updatecli.d/git-lfs.yml
+++ b/updatecli/updatecli.d/git-lfs.yml
@@ -11,7 +11,6 @@ sources:
       token: "{{ requiredEnv .github.token }}"
       username: "{{ .github.username }}"
       versionFilter:
-        pattern: ~2
         kind: semver
   # Retrieve the line defining Git-LFS version from Packer Auto Vars. file (it's a trick to handle partial file update)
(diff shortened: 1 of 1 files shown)


olblak (approved):

timja (approved):

olblak:
It appears that tests must passed before moving on

updatecli/updatecli.d/git-lfs.yml
```diff
@@ -11,7 +11,6 @@ sources:
       token: "{{ requiredEnv .github.token }}"
       username: "{{ .github.username }}"
       versionFilter:
-        pattern: ~2
```
dduportal, line 14: Either you have to update this line and set it to `pattern: ~3` to ensure that the git-lfs version if always 3.x (and avoid breaking things when the 4.x we'll be there), either you should keep it deleted but set the attribute `kind` to `latest`

dduportal (changes requested):

PR #162, opened 2022-01-10
moving packer fmt -recursive

smerle33:
just move fmt from packer file to script.


dduportal (approved):
LGTM: the only error is still jq

PR #173, opened 2022-01-28
Remove curl from ubuntu-20-provision.sh

smerle33:
following #172 I am removing curl as it's not needed anymore.

### DIFF
--- a/scripts/ubuntu-20-provision.sh
+++ b/scripts/ubuntu-20-provision.sh
@@ -62,7 +62,6 @@ function clean_apt() {
 function install_docker() {
   apt-get install -y --no-install-recommends \
     ca-certificates \
-    curl \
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

dduportal:
Mmmh why would we want to remove `curl` while it is useful on the agents?

smerle33:
> 

I would think that was only for the docker install and that if needed for the agents works, that would be installed elsewhere than here... what would be the process to know the tools used as the agent workload ?

dduportal:
> 

There isn't any process alas, except "mess around and find out when user complains" 😅

So: this PR was a good idea (cleaning up is always good), but bad luck we doàn't have a test suite to hold such "listing".

PR #182, opened 2022-02-28
chore: use `BRANCH_IS_PRIMARY` instead of `branch = 'main'`

lemeurherve:


timja (approved):

Jenkinsfile_k8s
```diff
@@ -123,9 +123,9 @@ pipeline {
           PACKER_GITHUB_API_TOKEN       = credentials('updatecli-github-token')
         }
         stages {
-          stage('Prepare on Branch Main') {
+          stage('Prepare on primary branch') {
             when {
-              branch 'main'
+              expression { env.BRANCH_IS_PRIMARY }
```
lemeurherve, line 128:
```suggestion
              expression { (env.BRANCH_IS_PRIMARY).toBoolean() }
```
env.BRANCH_IS_PRIMARY is set to `true` or unset, but we can't be sure its value hasn't been modified/set elsewhere.
Casting here to boolean ensure the `expression` function wouldn't interpret it as "existing", and thus executing it.

L2 and L43 don't have this problem as the "false" value of the env var will be correctly interpreted.

Jenkinsfile_k8s
```diff
@@ -123,9 +123,9 @@ pipeline {
           PACKER_GITHUB_API_TOKEN       = credentials('updatecli-github-token')
         }
         stages {
-          stage('Prepare on Branch Main') {
+          stage('Prepare on primary branch') {
             when {
-              branch 'main'
+              expression { (env.BRANCH_IS_PRIMARY).toBoolean() }
```
lemeurherve, line 128:
```suggestion
              expression { env.BRANCH_IS_PRIMARY.toBoolean() }
```

Jenkinsfile_k8s
```diff
@@ -123,9 +123,9 @@ pipeline {
           PACKER_GITHUB_API_TOKEN       = credentials('updatecli-github-token')
         }
         stages {
-          stage('Prepare on Branch Main') {
+          stage('Prepare on primary branch') {
             when {
-              branch 'main'
+              expression { env.BRANCH_IS_PRIMARY.toBoolean() }
```
lemeurherve, line 128:
```suggestion
              expression { env.BRANCH_IS_PRIMARY }
```

image

Issue #183, opened 2022-03-01
Use an azure credential instead of multiple string credentials to configure the Service Principal


lemeurherve:
https://github.com/jenkins-infra/packer-images/blob/main/Jenkinsfile_k8s#L114-L116

It would be nice to use an azure credential regrouping these 3 values.
Might need an additional plugin.

dduportal:
@timja as per https://plugins.jenkins.io/azure-credentials/, we can use the `withCredential`. Do you know if using the declarative form behaves with the "default" behavior described in https://plugins.jenkins.io/azure-credentials/#plugin-content-using-azurecredentials-in-a-job-freestyle--pipeline?

Eg.. how would the following snippet behave?

```
environment {
  AZURE_CLIENT = credentials('my-super-azure-credential')
}
```

?

dduportal:
Self answer: just tried it and it does not work with the credential form:

```
18:57:06  ERROR: No suitable binding handler could be found for type com.microsoft.azure.util.AzureCredentials. Supported types are StandardUsernamePasswordCredentials,FileCredentials,StringCredentials,DockerServerCredentials,AmazonWebServicesCredentials,SSHUserPrivateKey.
```

Worth a feature request to the plugin?

timja:
Sure, it works on scripted but maybe not declarative

not sure how you would model that

dduportal:
> Sure, it works on scripted but maybe not declarative
> 
> not sure how you would model that

I was thinking something like the `_USR` and `_PSW` trick for the usernamePassword and sshCredentials (ref. https://www.jenkins.io/doc/book/pipeline/syntax/#environment), reusing the same pattern as what the azure-credentials already does in scripted

Something like

```
environment {
  MY_CRED = credentials('my-super-azure-credential')
}

stages {
  stage('build') {
    steps {
      echo "My client id is $MY_CRED_CLIENT_ID"
      echo "My client secret is $MY_CRED_CLIENT_SECRET"
      echo "My tenant id is $MY_CRED_TENANT_ID"
      echo "My subscription id is $MY_CRED_SUBSCRIPTION_ID"
    }
  }
}

```

timja:
could you create an issue please?

dduportal:
> could you create an issue please? https://github.com/jenkinsci/azure-credentials-plugin

Here you are: https://github.com/jenkinsci/azure-credentials-plugin/issues/96 :) 
[...]

PR #185, opened 2022-03-08
chore: GC for azure resources group from packer process

smerle33:
Find every resources group that start with "pkr-Resource-" and is more than 1 day old from the tag timestamp and delete them as per the model for AWS : https://github.com/jenkins-infra/helpdesk/issues/2812

2 differences:  

-    no --dry-run in azure so I switch to the show sub-command to dry run
-    no batch delete command so I loop within the result list to delete one by one.   

PS : didn't add a location check as per : https://github.com/jenkins-infra/packer-images/blob/7a990b12227136625752012755dda690dc1475c2/jenkins-agent.pkr.hcl#L179 this seems to always be "eastus"

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -37,7 +37,7 @@ pipeline {
         stage('GC on Azure') {
           environment {
             PACKER_AZURE = credentials('packer-azure-serviceprincipal')
-            DRYRUN = "${env.BRANCH_NAME == 'main' ? 'false' : 'true'}"
+            DRYRUN = "${env.BRANCH_IS_PRIMARY ? 'false' : 'true'}"
           }
           
           agent {
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -36,7 +36,6 @@ pipeline {
         }
         stage('GC on Azure') {
           environment {
-            
             PACKER_AZURE = credentials('packer-azure-serviceprincipal')
             DRYRUN = "${env.BRANCH_NAME == 'main' ? 'false' : 'true'}"
           }
@@ -49,8 +48,8 @@ pipeline {
           
           steps {
             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
-              sh 'az login --service-principal -u "$AZURE_CLIENT_ID" -p "$AZURE_CLIENT_SECRET" -t "$AZURE_TENANT_ID"'
-              sh 'az account set -s "$AZURE_SUBSCRIPTION_ID"'
+              sh 'az login --service-principal -u "$PACKER_AZURE_CLIENT_ID" -p "$PACKER_AZURE_CLIENT_SECRET" -t "$PACKER_AZURE_TENANT_ID"'
+              sh 'az account set -s "$PACKER_AZURE_SUBSCRIPTION_ID"'
               sh './cleanup/azure.sh'
             }
           }
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -36,9 +36,8 @@ pipeline {
         }
         stage('GC on Azure') {
           environment {
-            PKR_VAR_azure_subscription_id = credentials('packer-azure-subscription-id')
-            PKR_VAR_azure_client_id       = credentials('packer-azure-client-id')
-            PKR_VAR_azure_client_secret   = credentials('packer-azure-client-secret')
+            
+            PACKER_AZURE = credentials('packer-azure-serviceprincipal')
             DRYRUN = "${env.BRANCH_NAME == 'main' ? 'false' : 'true'}"
           }
           
@@ -50,7 +49,7 @@ pipeline {
           
           steps {
             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
-              sh 'az login --service-principal -u "$PKR_VAR_azure_client_id" -p "$PKR_VAR_azure_client_secret"'
+              sh 'az login --service-principal -u "$AZURE_CLIENT_ID" -p "$AZURE_CLIENT_SECRET" -t "$AZURE_TENANT_ID"'
               sh 'az account set -s "$AZURE_SUBSCRIPTION_ID"'
               sh './cleanup/azure.sh'
             }
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -34,6 +34,32 @@ pipeline {
             }
           }
         }
+        stage('GC on Azure') {
+          environment {
+            PKR_VAR_azure_subscription_id = credentials('packer-azure-subscription-id')
+            PKR_VAR_azure_client_id       = credentials('packer-azure-client-id')
+            PKR_VAR_azure_client_secret   = credentials('packer-azure-client-secret')
+            DRYRUN = "${env.BRANCH_NAME == 'main' ? 'false' : 'true'}"
+          }
+          
+          agent {
+            kubernetes {
+              yamlFile 'CiPodTemplate.yaml'
+            }
+          }
+          
+          steps {
+            catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
+              sh '''
+                az login --service-principal \
+                    -u "$PKR_VAR_azure_client_id" \
+                    -p "$PKR_VAR_azure_client_secret" \
+                    -t "$PKR_VAR_azure_subscription_id"
+              '''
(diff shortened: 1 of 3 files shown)


cleanup/azure.sh
```diff
@@ -0,0 +1,63 @@
...
+    # Execute command "as it"
+    echo "command that would be executed : "
+    echo "az group" "$@"
+    az group "$@"
+  else
+    # Execute command with the "--dry-run"
+    subcommand="$1"
+    shift
+    echo "== DRY RUN (show instead of ${subcommand}):"
+    echo "az group show $@"
```
dduportal, line 18: Shellcheck:

```
In cleanup/azure.sh line 18:
    echo "az group show $@"
                        ^-- SC2145 (error): Argument mixes string and array. Use * or separate argument.
```

cleanup/azure.sh
```diff
@@ -0,0 +1,63 @@
...
+INSTANCE_IDS="$(az group list --query '[?tags.timestamp<='\''`'"${yesterday}"'`'\''] | [?starts_with(name, '\''pkr-Resource-'\'')].name'| jq -r '.[]' | xargs)" 
+
+if [ -n "${INSTANCE_IDS}" ]
+then
+  #shellcheck disable=SC2086
+  #has to run on each resource group
+  cpt="0"
+  for rg in ${INSTANCE_IDS}
+  do
+    run_az_deletion_command delete --name $rg
```
dduportal, line 55: Shellcheck:

```
In cleanup/azure.sh line 55:
    run_az_deletion_command delete --name $rg
                                          ^-^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
    run_az_deletion_command delete --name "$rg"
```

dduportal (changes requested):
Thanks for this!

For the first "quick & superficial" (e.g. haven't check the script deep dive) review, I got 2 blockings points:

- 2 shellchecks alerts
- It sounds like that the script is not executed by the pipeline: can you update the `Jenkinsfile_k8s` so that it execute the GC on Azure in parallel with the current AWS GC please?

cleanup/azure.sh
```diff
@@ -0,0 +1,63 @@
...
+    echo "command that would be executed : "
+    echo "az group" "$@"
+    az group "$@"
+  else
+    # Execute command with the "--dry-run"
+    subcommand="$1"
+    shift
+    echo "== DRY RUN (show instead of ${subcommand}):"
+    echo "az group show $@"
+    az group show "$@" 2>&1 | grep -v 'Request would have succeeded' || true
```
dduportal, line 19: 🤔 I'm not sure to understand why the command here is not the same as the one on line 11 (in the "real deletion" case branch)?

In the AWS script, there is a bunch of argument manipulation because aws CLI provides a `--dry-run` flag. Alas the `az` CLI does not.

I have 2 minor annoyances with the current behavior:
- By using the `az group show`, the output show a lot of verbose information that does not feel useful: I would expect only the resource groups names (and/or IDs) to be shown in dry run.
- When using dry run (by default) it ain't easy to really determine the command that should be run in production once deletion time is there.


WDYT about, instead of

```
== DRY RUN (show instead of delete):
az group show --name pkr-Resource-Group-zzxpwiae3b
{
  "id": "/subscriptions/dff2ec18-6a8e-405c-8e45-b7df7465acf0/resourceGroups/pkr-Resource-Group-zzxpwiae3b",
  "location": "eastus",
  "managedBy": null,
  "name": "pkr-Resource-Group-zzxpwiae3b",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "build_type": "dev",
    "imageplatform": "amd64",
    "imagetype": "jenkins-agent-windows-2019",
    "scm_ref": "b14bece",
    "timestamp": "20220127170111",
    "version": "0.0.1"
  },
  "type": "Microsoft.Resources/resourceGroups"
}
== DRY RUN (show instead of delete):
az group show --name pkr-Resource-Group-8dntm8cpki
{
  "id": "/subscriptions/dff2ec18-6a8e-405c-8e45-b7df7465acf0/resourceGroups/pkr-Resource-Group-8dntm8cpki",
  "location": "eastus",
  "managedBy": null,
  "name": "pkr-Resource-Group-8dntm8cpki",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "build_type": "dev",
    "imageplatform": "amd64",
    "imagetype": "jenkins-agent-windows-2019",
    "scm_ref": "c9d27b9",
    "timestamp": "20220211184117",
    "version": "0.0.1"
  },
  "type": "Microsoft.Resources/resourceGroups"
}
```


that we would have something like:

```
== DRY RUN
= Command that would be executed without dry-run:
az group delete --name pkr-Resource-Group-zzxpwiae3b --yes
= Command that would be executed without dry-run:
az group delete --namepkr-Resource-Group-8dntm8cpki
```
?

dduportal, line 19:
```suggestion
    az group show "$@" || true
```

not needed since the grep was an assertion based on the `aws` CLI

cleanup/azure.sh
```diff
@@ -0,0 +1,63 @@
...
+timeshift_days="1"
+if date -v-${timeshift_days}m > /dev/null 2>&1; then
+    # BSD systems (Mac OS X)
+    yesterday="$(date -v-${timeshift_days}d +%Y%m%d%H%M%S)"
+else
+    # GNU systems (Linux)
+    yesterday="$(date --date="-${timeshift_days} days" +%Y%m%d%H%M%S)"
+fi
+
+## Check for aws API reachability (is it configured?)
```
dduportal, line 40:
```suggestion
## Check for Azure API reachability (is it configured?)
```

cleanup/azure.sh
```diff
@@ -0,0 +1,63 @@
...
+az account show >/dev/null || \
+  { echo "[ERROR] Unable to request the Azure API: the command 'az account show' failed. Please check your Azure credentials"; exit 1; }
+
+## Remove running instances older than 24 hours
+#az group list --query '[?tags.timestamp<'\''20220222222222'\''] | [?starts_with(name, '\''pkr-Resource-'\'')].name'
```
dduportal, line 45:
```suggestion
```

Dangling comment? :)

cleanup/azure.sh
dduportal, line 55:
```suggestion
    run_az_deletion_command delete --name $rg --yes
```

Non interactive output

dduportal:
> no batch delete command so I loop within the result list to delete one by one.

WDYT about adding the flag `--no-wait` to the az delete command, so that the loop is not stuck waiting for a resource to be deleted? It would mean asynchronous deletion.

cleanup/azure.sh
smerle33: 👍🏻 love it

smerle33: note needed anymore by the way it should handle the dry run now

cleanup/azure.sh
smerle33: forgot this one thanks.

Jenkinsfile_k8s
```diff
@@ -34,6 +34,28 @@ pipeline {
             }
           }
         }
+        stage('GC on Azure') {
+          environment {
+            PKR_VAR_azure_subscription_id = credentials('packer-azure-subscription-id')
+            PKR_VAR_azure_client_id       = credentials('packer-azure-client-id')
+            PKR_VAR_azure_client_secret   = credentials('packer-azure-client-secret')
```
lemeurherve, line 41: If you have the time, take a look at https://github.com/jenkins-infra/packer-images/issues/183 to use an azure credentials instead.

(Could be done in another PR)

dduportal: will be done in another PR, but you are correct :)

Jenkinsfile_k8s
```diff
@@ -34,6 +34,26 @@ pipeline {
             }
           }
         }
+        stage('GC on Azure') {
+          environment {
+            PACKER_AZURE = credentials('packer-azure-serviceprincipal')
+            DRYRUN = "${env.BRANCH_NAME == 'main' ? 'false' : 'true'}"
```
lemeurherve, line 40:
```suggestion
            DRYRUN = "${env.BRANCH_IS_PRIMARY ? 'false' : 'true'}"
```

lemeurherve (approved):
🎉

dduportal (approved):
LGTM 👏

PR #193, opened 2022-03-18
chore(Jenkinsfile) Use githubApp credentials for packer instead of a Github bot user token

### LINKED ISSUE #189: Use githubApp credentials for packer instead of a Github bot user token
https://github.com/jenkins-infra/packer-images/blob/a876f3e2eaf731e1036feefcfe2a80f882fe60c2/Jenkinsfile_k8s#L141

Related: https://github.com/jenkins-infra/helpdesk/issues/2817

dduportal:
Closes #189 .

This PR switches from a github (personal) token usage to a githubapp => "Installation token usage".

Please note that the githubapp infra is used (instead of the updatecli githubapp) as packer init does NOT need any write access.

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -135,7 +135,7 @@ pipeline {
           PACKER_PLUGIN_PATH            = "${PACKER_HOME_DIR}/plugins"
           PACKER_VARS_FILE              = ".auto.pkrvars.hcl"
           // Reuse the updatecli Github token to ensure that packer init does not hit the rate limit
-          PACKER_GITHUB_API_TOKEN       = credentials('updatecli-github-token')
+          PACKER_GITHUB_API_TOKEN       = credentials('github-app-infra')
         }
         stages {
           stage('Prepare on primary branch') {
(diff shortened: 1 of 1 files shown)


timja (approved):
Could add a step to the shared library which gets the credential ID that was used to checkout the repo?

https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/checkoutScm.groovy#L24

dduportal:
> Could add a step to the shared library which gets the credential ID that was used to checkout the repo?
> 
> https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/checkoutScm.groovy#L24

Good idea. The only issue is that we have to find the correct syntax that would not need security approval if it is in a Jenkinsfile (as shared-library like you linked is allowed to approve methods such a rawBuild).


But right now, the builds says HTTP/401. We have to check if there is a valid IAT generated into the variable and check why this IAT is not allowed (I assume it might be related to github app installation scopes)?

lemeurherve (approved):

timja:
> Good idea. The only issue is that we have to find the correct syntax that would not need security approval if it is in a Jenkinsfile (as shared-library like you linked is allowed to approve methods such a rawBuild).

That's why I said shared library :) , I'm not aware of a syntax that will work without that.

> But right now, the builds says HTTP/401. We have to check if there is a valid IAT generated into the variable and check why this IAT is not allowed (I assume it might be related to github app installation scopes)?

👍  Likely scope related

PR #201, opened 2022-04-06
feat(GC AWS AMI): Garbage collector for DEV images

smerle33:
as per https://github.com/jenkins-infra/ideas/issues/36
remove the OLD AWS AMI marked as dev (more than 1 month old).
to be continued with : 

- [ ]  stagging
- [ ] prod 

but need more input as for the filtering


dduportal (changes requested):
don't forget to add the script to the pipeline (otherwise it will not be run :) )

smerle33:
> don't forget to add the script to the pipeline (otherwise it will not be run :) )

I did on the second commit ... just before your review ...😎

Issue #198, opened 2022-03-31
Add jenkins-infra tools


dduportal:
The jenkins infrastructure requires additional tools that should be installed in these templates:

- [x] The [`jx-release-version`](https://github.com/jenkins-x-plugins/jx-release-version) because of https://github.com/jenkins-infra/pipeline-library/blob/544d04d1dbc67cbe7aed681924c44488bb00b6b6/vars/buildDockerAndPublishImage.groovy#L63-L74
- [x] The [`gh`](https://github.com/cli/cli) CLI because of https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildDockerAndPublishImage.groovy#L178-L189
- [x] The [`container-structure-test`](https://github.com/GoogleContainerTools/container-structure-test) tool because of https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildDockerAndPublishImage.groovy#L110-L129
~- [ ] The ['uc`](https://github.com/jenkins-infra/uc) and [`jenkins-version`](https://github.com/jenkins-infra/jenkins-version) CLIs because of https://github.com/jenkins-infra/docker-jenkins-weekly and https://github.com/jenkins-infra/docker-jenkins-lts~ (removed from initial list as we favor using the jenkins-plugin-cli instead)
- [x] The [`updatecli`](https://github.com/updatecli/updatecli) because... well... jenkins infra uses it everywhere - https://github.com/jenkins-infra/packer-images/pull/398
- [x] The [`hadolint`](https://github.com/hadolint/hadolint) tool because of https://github.com/jenkins-infra/pipeline-library/blob/544d04d1dbc67cbe7aed681924c44488bb00b6b6/vars/buildDockerAndPublishImage.groovy
- [x] The ['netlify-golang-deploy`](https://github.com/halkeye/netlify-golang-deploy) CLI by @halkeye because of https://github.com/jenkins-infra/plugin-site/blob/0df7d8e482ba9bc38492a324f6d1f64c2daa4b6d/Jenkinsfile_k8s#L183, https://github.com/jenkins-infra/jenkins-is-the-way/blob/ba65a589f7a44de2d542026852f22d99dd62eeed/Jenkinsfile_k8s#L155 and https://github.com/jenkins-infra/jenkins.io/blob/65ab9b3afea35bcef7128a229e6f1a34158223f1/Jenkinsfile_k8s#L122
  => https://github.com/jenkins-infra/packer-images/pull/404
[...]

PR #216, opened 2022-04-29
feat: add a first batch of infra tools

dduportal:
This PR is a firt batch of infra tools installation as per #198.

It helps on the the following issues:

- https://github.com/jenkins-infra/repository-permissions-updater/issues/2520 to fix the RPU failing build missing the `az` CLI on the VM agents
-  https://github.com/jenkins-infra/helpdesk/issues/2873 to provide all the expected Docker tooling on both Linux and Windows.

It also fixes the Windows templates to ensure that the installed tools are available.

In details:

- chores:
  - Switch the `.autopkrvar.hcl` logic (packer variables with implicit default values) to an external YAML file loaded dynamically: no need to declare in packer each provisioning environment variable expressing tool versions.
    - Note that this change breaks updatecli diff stage (as usual) on this PR but will work once merged
  - Fix the pipeline interpolation warnings (to avoid the risk around credentials)
  - Add a set of "sanity checks" on each platform, executed at the end of provisionning, printing out all the command line tools versions and installed packages. This check is helpful to underline issues when a command line is not on the correct location, is not executable, not named correctly (`.exe` suffix is mandatory on CLI on windows...)
  - Updated/added all updatecli manifests for all tools touched by this change request to use the new YAML + remove updatecli  manifest warnings
- Features:
  - Add `azure-cli` 2.36.0
  - Add `gh ` (GitHub CLI) 2.39.0
  - Add `hadolint` 2.10.0
  - Add `container-structure-test` 1.11.0
  - Add `jx-release-version` 2.5.1
  - Bump docker-compose to `2.5.0` (from `1.29.2`) and make it available on ALL platforms (Windows included)
  - Fix `jq` installation on all platform + pin it to `1.6`


scripts/ubuntu-20-provision.sh
```diff
@@ -191,6 +193,56 @@ function install_maven() {
...
 }
 
+## Ensure that hadolint is installed
+function install_hadolint() {
+  hadolint_arch_suffix="${ARCHITECTURE}"
+  if [ "${hadolint_arch_suffix}" == "amd64" ]; then
+    hadolint_arch_suffix="x86_64"
+  fi
+  curl --fail --silent --location --show-error --output /usr/local/bin/hadolint \
+    "https://github.com/hadolint/hadolint/releases/download/v${HADOLINT_VERSION}/hadolint-Linux-${hadolint_arch_suffix}"
```
lemeurherve, line 203: Does the same method as L182 for docker-compose be used here too?
If so,
```suggestion
  curl --fail --silent --location --show-error --output /usr/local/bin/hadolint \
    "https://github.com/hadolint/hadolint/releases/download/v${HADOLINT_VERSION}/hadolint-Linux-$(uname -m)"
```
(I suspect there is a reason you didn't do that)

scripts/windows-2019-provision.ps1
```diff
@@ -215,3 +250,23 @@ Select-Object -Property DeviceID, DriveType, VolumeName,
...
+& "$baseDir\git\cmd\git.exe" --version
+& "$baseDir\jdk-8\bin\java.exe" -version
+& "$baseDir\jdk-11\bin\java.exe" -version
+& "$baseDir\jdk-17\bin\java.exe" -version
+& "$baseDir\container-structure-test.exe" version
+& "$baseDir\hadolint.exe" --version
+& "$baseDir\docker-compose.exe" --version
+& "$baseDir\git\mingw64\bin\git-lfs.exe" --version
+& "$baseDir\jx-release-version" --version
+& "$baseDir\jq.exe" --version
```
lemeurherve, line 266: Is the $baseDir added to the $PATH? If so, shouldn't the sanity checks run without the folder to ensure they're available everywhere?

scripts/windows-2019-provision.ps1
```diff
@@ -215,3 +250,23 @@ Select-Object -Property DeviceID, DriveType, VolumeName,
...
+& "C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\wbin\az.cmd" version
```
lemeurherve, line 267: Same question, is this folder included in the $PATH?
Or should consumers of the tool know where they are stored?

lemeurherve (commented):
Great job!

A minor question around architecture in linux links, and another around the $PATH, otherwise LGTM

scripts/ubuntu-20-provision.sh
dduportal: You suspect well ;) 

The output of `uname -m` on ARM64 CPUs is: `aarch64`

```bash
$ docker run --rm -ti --platform=linux/arm64 --entrypoint=uname ubuntu:20.04 -m 
aarch64
```

So problem here is that each tool have it's own usage (Ouptut of `GOOS` and `GOARCH` - https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63, output of `uname -m`, custom like what we are doing, etc.)

OItherwise I would have used the same variable everywhere

scripts/windows-2019-provision.ps1
dduportal: That is the pattern with shell script yes.

But with powershell, the PATH is written in the local register, and is read only after the VM reboot if I understand correctly.

There might be a technique that I don't know about, but even when I failed with trying to spawn either a new `cmd` or `powershell` subprocess, so decided to go that way: at least it allowed be to discover issues on the current Windows templates :)

scripts/windows-2019-provision.ps1
dduportal: Same answer as the previous (good) question for the PATH.

In the case of the `azure-cli`, it's the MSI installation that decides of this default installation.
Did not check if I could customize the installation path: this "quick and dirty" absolute path is enough for the goal of the test: WDYT?

dduportal:
Note: the build is OK, only the `updatecli diff` stage fails with the message below for each failed target:

```
TARGETS
========
updateMavenVersion
------------------
**Dry Run enabled**
ERROR: Something went wrong in target "updateMavenVersion" :
ERROR: the yaml file "/tmp/updatecli/jenkins-infra/packer-images/provision-env.yml" does not exist

PULL REQUESTS
=============
ERROR: 1 target(s) (updateMavenVersion) failed for pullrequest "default"
```

You can proceed to merge if it is ok for you dear reviewers :)

PR #219, opened 2022-05-05
feat(windows) bump Windows Server to 2022 edition

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2629.

Depends on #221 and #239.

This PR introduces an additional set of VM templates with Windows Server Core 2022 as base OS.

Please note that:
- Docker container is excluded (for now): only AWS and Azure VM templates are added.
- Windows Server 2019 is kept (see comment below for details)

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -87,7 +87,7 @@ pipeline {
           }
           axis {
             name 'agent_type'
-            values 'ubuntu-20.04', 'windows-2019'
+            values 'ubuntu-20.04', 'windows-2019', 'windows-2022'
           }
           axis {
             name 'compute_type'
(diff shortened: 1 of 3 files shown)


dduportal:
As per the last SIG platform meeting, it seems that this change request should be changed to build both Server 2022 and Server 2019.

Reason is that, as per https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2022%2Cwindows-11-21H2, running Windows Server 2019 containers on a 2022 host would result in using Hyper-V nested virtualization.
Performances might be impacted so keeping 2 lines of Windows server at first, until 2019 Docker images are deprecated, is a much bug-prone road...


Thanks @timja and @MarkEWaite for this notif !

dduportal:
ToDo list:

- [ ] Add the EC2launcher configuration v2 in the provisionning powershell script (instead of packer inline) to 2019 only + check if 2022 needs it
- [ ] Select an Azure base image with Docker engine for Windows 2022 or install Docker engine

timja (approved):

PR #247, opened 2022-05-30
fix: install git for Windows with choco so we can chose to not install Git Credential Manager

lemeurherve:
Related: https://github.com/jenkins-infra/helpdesk/issues/2873


lemeurherve:
Superceded by https://github.com/jenkins-infra/packer-images/pull/248

PR #248, opened 2022-05-31
fix: no cygwin

lemeurherve:
It seems there is a collision between them and the tools included with git for Windows when used in the buildDockerAndPublishImage shared pipeline lib:

> Cloning into 'private-repo'...
>       0 [main] cat (2272) C:\tools\git\usr\bin\cat.exe: *** fatal error - cygheap base mismatch detected - 0x180348408/0x18034C408.
> This problem is probably due to using incompatible versions of the cygwin DLL.

Related: https://github.com/jenkins-infra/helpdesk/issues/2873


dduportal (approved):
🚀

dduportal (approved):

PR #250, opened 2022-05-31
chore(ubuntu) switch git-lfs installation to packagecloud.io packages

dduportal:
Ref. https://github.com/jenkinsci/docker/pull/1379


dduportal:
Closing as it fails.

Reason: we are using Ubuntu 20.04 (focal), which is not supported by package cloud (only jammy and impish).

Why does it work on https://github.com/jenkinsci/docker/pull/1379 ? It's because the Jenkins Docker official images is on Debian bullseye.

PR #272, opened 2022-07-11
fix(linux) correct installations of asdf and ruby for user `jenkins`

dduportal:
In #269 asdf was installed into `/opt/asf` (system-wide).

Even though supported, it is not standard and makes the permissions management complex.
Also, the Ruby installation with ASDF was done in the context of the root user making it unusable in our agents running with the (unprivileged) user `jenkins`.

This PR fixes both issues:
- ASDF is now installed for the user `jenkins` following the official ASDF documentation (Bash & Git)
- Introducing 2 new functions to add ASDF plugins and packages (because a file need to be sourced to load ASDF in current PATH)
- Ruby is installed with these 2 new functions, in `jenkins` user context (not system wide)
- Reordering functions to fulfill ASDF requirements
- Updating the sanity check to run as `jenkins` user


provisioning/ubuntu-provision.sh
```diff
@@ -64,6 +66,72 @@ function clean_apt() {
...
+
+## Install the ASDF Plugin passed as argument ($1 is the name and $2 the URL)
+function install_asdf_plugin() {
+  local plugin_name="${1}"
+  local plugin_url="${2}"
+
+  # Git is required to install asdf plugins
+  command -v git >/dev/null 2>&1
+
+  # sleep 10000
```
timja, line 119: ?

provisioning/ubuntu-provision.sh
```diff
@@ -82,6 +150,10 @@ function install_docker() {
     $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
   apt-get update --quiet
   apt-get install --yes --no-install-recommends docker-ce
+
+  # Allow the default user to use Docker. https://docs.docker.com/engine/install/linux-postinstall/
+  # Please note that it gives effcitvely full root permissions to this user so these compute instances must ephemeral
```
timja, line 155:
```suggestion
  # Please note that it gives effcitvely full root permissions to this user so these compute instances must be ephemeral
```

provisioning/ubuntu-provision.sh
dduportal: ah ah you caught me red-handed: forgot to remove my debug stances. Thanks Tim.

dduportal, line 119:
```suggestion
```

PR #286, opened 2022-07-20
chore: add a retry on the pipeline

### LINKED ISSUE #284: Add a Retry on the pipeline
Quite often, the Azure Packer provider crashes at the end of the process, when trying to stop the Azure VM right before exporting it to an image.

It is a transient error: we trigger manually a new build and it always ends up on success after 2 or 3 tries.

We should us the `retry` pipeline logic to stop caring about manually re-triggering.

Downside: retrying will cross the timeout limit: need to increase?

lemeurherve:
Closes #284

Let's see if retried builds are timed out before increasing [its default value](https://github.com/jenkins-infra/packer-images/blob/main/Jenkinsfile_k8s#L24).

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -164,9 +164,6 @@ pipeline {
         stages {
           stage('Build Template') {
             steps {
-              def attempt = 0
-              def attempts = 2
-
               // Pass the Azure credentials through Packer Input variables, as environment variables prefixed with 'PKR_VAR_'
                 // Re.f https://www.packer.io/docs/templates/hcl_templates/variables#assigning-values-to-build-variables
               withCredentials([azureServicePrincipal(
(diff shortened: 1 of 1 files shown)


Jenkinsfile_k8s
```diff
@@ -192,7 +192,12 @@ pipeline {
                 sh 'packer validate ./'
 
                 // Execute build only for this matrix cell's setup
-                sh 'packer build -timestamp-ui -force -only="${PKR_VAR_image_type}.${PKR_VAR_agent_os_type}" ./'
+                def attempt = 0
+                def attempts = 2
+                retry(count: attempts, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+                  echo '== Attempt ' + ++attempt + ' of ' + attempts
```
dduportal, line 198: How is the `attempt` variable incremented?

lemeurherve: From the `++attempt`

dduportal: oh I see 👍

dduportal (approved):

PR #287, opened 2022-07-22
feat: release Docker image for Linux

lemeurherve:
Ref: https://github.com/jenkins-infra/packer-images/issues/282

Issue #282, opened 2022-07-20
Release Docker Image for Linux (Ubuntu 20.04)


dduportal:
- Add a specific provisioning only with the `docker.ubuntu` source, which download the `agent.jar` and the `entrypoint.sh` (at least?) 
- Add the required Docker Image metadatas (`USER`, `ENTRYPOINT`, etc.) from the actual `jenkinsci/inbound-agent` in the Packer source configuration `docker.ubuntu` along with enabling the `commit` feature
- Add a `docker-tag` post-processor in packer (to give a tagged name to the local Docker image)
- Add a **conditional** `docker-push` post-processor in packer (only if the variable `build_type` has the value `prod`) to ensure that we only push images when the packer builds are triggered by a new tag

dduportal:
Closed by #287 #291

PR #291, opened 2022-07-26
fix: release Docker image for Linux

lemeurherve:
Follow-up of https://github.com/jenkins-infra/packer-images/pull/287

Use correct credentials (needs https://github.com/jenkins-infra/kubernetes-management/pull/2644)
Set isPrimary variable as env.BRANCH_IS_PRIMARY doesn't seem to be available in scripted
Add a condition on compute_type for the 'publish docker image' stage

Ref: https://github.com/jenkins-infra/packer-images/issues/282

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -1,6 +1,6 @@
-def isPrimary = false
+def isPrimaryBranch = false
 if (env.BRANCH_IS_PRIMARY) {
-  isPrimary = true
+  isPrimaryBranch = true
   properties([
     buildDiscarder(logRotator(numToKeepStr: '10')),
     // Daily build is enough: only the tagged build would generate downstream PRs on jenkins-infra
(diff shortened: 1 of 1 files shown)


Jenkinsfile_k8s
```diff
@@ -1,5 +1,6 @@
-
+def isPrimary = false
```
dduportal, line 1:
```suggestion
def isBranchPrimary = false
```

WDYt about this name proposal? Rationale is that the qualificative "primary" could be confusing. By adding the name "branch", it aims to make explic when reading the name that "branch" is the subject of the qualificative

dduportal (approved):
👍

dduportal (approved):

Jenkinsfile_k8s
smerle33: I think I mistakenly clicked a button around here ... sorry ....

lemeurherve: @smerle33 which button? I don't see any kind of modification, should be fine!

PR #294, opened 2022-07-26
fix: correct call for docker push

lemeurherve:


Jenkinsfile_k8s
```diff
@@ -204,7 +204,7 @@ pipeline {
                   infra.withDockerPushCredentials {
                     def dockerImage = "jenkinsciinfra/jenkins-agent-${env.agent_type}:${env.TAG_NAME}"
                     echo "Publishing ${dockerImage}..."
-                    docker push dockerImage
+                    sh "docker push ${dockerImage}"
```
dduportal, line 207:
```suggestion
                    sh 'docker push jenkinsciinfra/jenkins-agent-${agent_type}:${TAG_NAME}'
```

- No groovy interpolation: only shell script reading its variables (I might be mistaken on the real name but anything in the `env` pipeline object is expected  to be serialized in the `sh` environment).
- Don't forget to remove the `def dockerImage` (as it is only used there)

dduportal (approved):

PR #308, opened 2022-08-05
chore(azure builders) enable async resources deletion

dduportal:
Ref. https://www.packer.io/plugins/builders/azure/arm#async_resourcegroup_delete

We have, quite often, error (go panic) thrown by the azure packer plugin.

This repository used to have this setting enabled (ref. https://github.com/jenkins-infra/packer-images/pull/61/files) but it was removed at a given point in time, because there was no clud resource GCs.


timja (approved):

dduportal:
Closing as I'm blind: this setting was already set up in the `sources.hcl`... 😅

PR #309, opened 2022-08-05
feat: add packer 1.8.2

dduportal:
- Uses ASDF on linux, to benefit from checskum scan with GPG integrity check
- Uses chocolatey on Windows for the same reason
- Added a condition check on the updatecli manifest to verify that the package is published on chocolatey (sometimes it takes 9-10 days for chocolatey)


dduportal:
Please note that, once deployed, this PR would allow to cleanup https://github.com/jenkins-infra/packer-images/blob/21931537821fd0f2de74fc6b5f6c74cc1d5a24e2/Jenkinsfile_k8s#L174-L175

PR #336, opened 2022-09-13
fix(datadog): disabling datadog at startup

smerle33:
to avoid bad logs before cloud-init generation of datadog.yaml

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -386,6 +386,7 @@ function install_datadog() {
 
   apt-get update
   apt-get install --yes --no-install-recommends datadog-agent datadog-signing-keys
+  systemctl disable datadog-agent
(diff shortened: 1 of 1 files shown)


provisioning/ubuntu-provision.sh
```diff
@@ -386,6 +386,7 @@ function install_datadog() {
 
   apt-get update
   apt-get install --yes --no-install-recommends datadog-agent datadog-signing-keys
+  systemctl disable datadog-agent
```
dduportal, line 389:
```suggestion
  # Disabling service at startupt because <....>
  systemctl disable datadog-agent
```

Do you mind adding a comment to explain "why" ? Otherwise LGTM

dduportal (approved):

dduportal (approved):

PR #340, opened 2022-09-15
fix (locale): set the locale

smerle33:


provisioning/ubuntu-provision.sh
```diff
@@ -66,6 +66,10 @@ function clean_apt() {
   apt-get upgrade -y
 }
 
+function setlocale() {
+  localectl set-locale LANG=en_US.UTF-8
```
dduportal, line 70: This command will not work on containers, because `systemd` and all its mess (`*ctl` commands, dbus, etc.) requires PID 1 and privileged permissions alas.

You might want to check https://help.ubuntu.com/community/Locale for alternatives (don't forget to generate the locale).

dduportal:
Closing as duplicate of #341

PR #346, opened 2022-09-16
chore(ubuntu provisioning) apt-get update

smerle33:
@dduportal is that more understandable ?

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -62,7 +62,7 @@ function clean_apt() {
   apt-get autoremove --purge -y
 
   ## Ensure the machine is up-to-date
-  apt-get update
+  apt-get update --quiet
   apt-get upgrade -y
 }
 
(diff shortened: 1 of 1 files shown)


dduportal (approved):
Oh yeah, that make sense, many thanks!

provisioning/ubuntu-provision.sh
```diff
@@ -62,7 +62,7 @@ function clean_apt() {
   apt-get autoremove --purge -y
 
   ## Ensure the machine is up-to-date
-  apt-get update
+  apt-get update --quiet
```
lemeurherve, line 65: Maybe create an `apt_update` function?
WDYT @dduportal @smerle33 ?
(nit)

dduportal: One line to factorize one line seems weird

dduportal:
Merging as the checks errors comes from Windows Azure VMs (golang panic at the end).

PR #389, opened 2022-10-23
Add chromium version 105.0.5195.102

timja:
Required for https://github.com/jenkins-infra/account-app/pull/167

see comments in code for some of the choices.

Unsure if it's worth adding to Windows, would rather not add it if it would never be used there

### DIFF
--- a/provisioning/tools-versions.yml
+++ b/provisioning/tools-versions.yml
@@ -1,6 +1,7 @@
 asdf_version: "0.10.2"
 azurecli_version: "2.41.0"
 chocolatey_version: "1.2.0"
+chromium_version: "105.0.5195.102"
 compose_version: "2.12.2"
 cst_version: "1.11.0"
 default_jdk: "11"
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -246,7 +246,7 @@ Pin: release o=LP-PPA-phd-chromium-browser
 Pin-Priority: 1001
 ' | tee /etc/apt/preferences.d/phd-chromium-browser
 
-  apt-get install --yes chromium-browser
+  apt-get install --yes chromium-browser="${CHROMIUM_VERSION}"-0ubuntu0.18.04.1
 
 }
 
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -226,6 +226,30 @@ function install_python() {
   python3 -m pip install --no-cache-dir pip --upgrade
 }
 
+## Install chromium
+### Using chromium because chrome is not available on arm64
+### see https://bugs.chromium.org/p/chromium/issues/detail?id=677140
+function install_chromium() {
+  apt-get remove chromium-browser chromium-browser-l10n chromium-codecs-ffmpeg-extra
+  apt-get update --quiet
+  apt-get install --yes software-properties-common
+
+  # chromium default repositories require snap which doesn't work in docker
+  # see https://askubuntu.com/questions/1255692/is-it-still-possible-to-install-chromium-as-a-deb-package-on-20-04-lts-using-som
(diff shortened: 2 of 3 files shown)


timja:
It's green now @jenkins-infra/core, not reporting back the last one to GitHub for some reason
image

timja:
possible for a review please @jenkins-infra/core?

dduportal:
> Required for https://github.com/jenkins-infra/account-app/pull/167

Side question: since Required for https://github.com/jenkins-infra/account-app/pull/167 uses VM, can you achieve what you want with a Docker image (temporarly)?

timja:
> (first superficial review: missing fixed version + updatecli, and missing on windows)

This may be possible with Chromium, it wasn't possible with Google Chrome as they remove old versions from the repository forcing people to stay up to date.

> missing on windows

Intentionally not added as no use-case for it, but if required can do so

> Side question: since Required for https://github.com/jenkins-infra/account-app/pull/167 uses VM, can you achieve what you want with a Docker image (temporarly)?

Could do but just complicates it and would rather keep it simpler with this.
I'm not absolutely blocked as I've got more tests to write, and can run them locally for now / disable them on CI.

dduportal:
> > (first superficial review: missing fixed version + updatecli, and missing on windows)
> 
> This may be possible with Chromium, it wasn't possible with Google Chrome as they remove old versions from the repository forcing people to stay up to date.
> 
> > missing on windows
> 
> Intentionally not added as no use-case for it, but if required can do so
> 
> > Side question: since Required for [jenkins-infra/account-app#167](https://github.com/jenkins-infra/account-app/pull/167) uses VM, can you achieve what you want with a Docker image (temporarly)?
> 
> Could do but just complicates it and would rather keep it simpler with this. I'm not absolutely blocked as I've got more tests to write, and can run them locally for now / disable them on CI.

We'll try our best but the work on account-app is not priority for infra so we need a bit of time for this. We'll try our best to unblock you here.

dduportal:
> This may be possible with Chromium, it wasn't possible with Google Chrome as they remove old versions from the repository forcing people to stay up to date.

Yep, we have the same issue with Git in the ubuntu PPA. At least it forces us to keep it up to date 😅 
If you can specify a fixed version, I'll take care of the updatecli manifest in an upcoming PR: looks good to you?

=> The goal is to have an auditable changelog for users (which reminds me that Docker should also be the same) because when there is a breaking change in dev tools, it is hard to audit today if version are not specified.

>>    missing on windows
> Intentionally not added as no use-case for it, but if required can do so

=> The idea is to get the same features on both OS. 

WDYT about using Chocolatey: https://community.chocolatey.org/packages?q=chromium (there is also chromium-stable)? It supports version (look at the other tools such as vagrant or ruby).
I'll also take care of the updatecli manifest for this one.

halkeye:
you could request a saucelabs open source license, and then not really worry about keeping dependencies for chrome.

dduportal (approved):

timja:
> you could request a saucelabs open source license, and then not really worry about keeping dependencies for chrome.

could do but I don't think we could run it on ci.jenkins.io which defeats a lot of the point.

provisioning/tools-versions.yml
```diff
@@ -1,6 +1,7 @@
 asdf_version: "0.10.2"
 azurecli_version: "2.41.0"
 chocolatey_version: "1.2.0"
+chromium_version: "105.0.5195.102"
```
timja, line 4: one point on this, the linux and windows available versions are currently different, windows is one version newer (although windows has old packages so it works fine).

something to keep in mind for updatecli.

could also just split them to windows / linux versions like in git

timja:
Thanks!

halkeye:
Sauce supports (unless they removed it) a limited time signed jwt token so it can be used in a public environment, and if the token leaks it's only good for like an hour and can only run tests not account APIs.

Not saying you should do it, just that it's an option

PR #395, opened 2022-11-03
chore: pin docker version for Linux and track it with updatecli

dduportal:
Signed-off-by: Damien Duportal <damien.duportal@gmail.com>


provisioning/ubuntu-provision.sh
```diff
@@ -183,6 +183,15 @@ function install_docker() {
     "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
     $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
   apt-get update --quiet
+
+  # Ensure that version is fixed
+  echo "
+Package: docker-ce*
+Pin: version 5:${DOCKER_VERSION}*
```
lemeurherve, line 190: What is the `5:` for?

dduportal: As far as I remember (to be searched and confirmed by the documentation), the package version prefix is the epoch. It is usually `0:` but in the case of `docker-ce` package it is `5: `. I assume it is some kind of trick to allow overriding at distribution level but I don't know really.

dduportal:
Putting back as a draft: the environment in which updatecli executes is already a container, so the command `docker` is not available.

dduportal:
```
The shell 🐚 command "bash ./updatecli/scripts/run-in-docker.sh ./updatecli/scripts/fetch-docker-ce-latest-version.sh" ran successfully with the following output:
---
20.10.21
--- 
```

=> looks good! I had to bypass the pipeline library: comment with link to the RFE added

PR #405, opened 2022-11-08
chore(updatecli) track versions of updatecli, awscli and netlify-deploy

dduportal:


updatecli/updatecli.d/awscli.yml
```diff
@@ -0,0 +1,55 @@
...
+      owner: "{{ .github.owner }}"
+      repository: "{{ .github.repository }}"
+      token: "{{ requiredEnv .github.token }}"
+      username: "{{ .github.username }}"
+      branch: "{{ .github.branch }}"
+
+sources:
+  lastReleaseVersion:
+    kind: githubrelease
+    name: Get the latest `aws` CLI` version
```
lemeurherve, line 19:
```suggestion
    name: Get the latest `aws` CLI version
```

updatecli/updatecli.d/awscli.yml
```diff
@@ -0,0 +1,55 @@
...
+    scmid: default
+
+pullrequests:
+  default:
+    kind: github
+    scmid: default
+    title: Bump the `aws` CLI version to {{ source "lastReleaseVersion" }}
+    spec:
+      labels:
+        - enhancement
```
lemeurherve, line 54: Just noticed it's "enhancement" here instead of "dependencies" like in the other repos.
WDYT if I open another PR to align this repo with the others?

lemeurherve (approved):

updatecli/updatecli.d/awscli.yml
dduportal: The initial idea was to separate "dependencies" (such as packer plugins, github actions) from "image dependency changes" (such as git version update for instance).

"Enhancement" means "should trigger a semantic version bump on the minor digit" while dependency is only about the tooling.

That is something that bother me a lot on the other repositories. But having a repo different than the others is clearly worse: go for your PR!
WDYT if in your PR you change the "depencencies" labels to "chore" (for the packer plugins) and the "enhancement" to "dependencies (and update the release-drafter config accordingly)?

PR #452, opened 2022-12-21
feat(Windows-AWS) Switch from EC2 Launch v1 to v2 (support of both 2019 and 2022)

dduportal:
This PR was initially aimed to fix Windows Server 2019 build errors on AWS, when EC2 Launch instructions were failing in December 2022.

The failure was caused by an error in the powershell provisioner for Windows AWS images, when running the [EC2Lauch](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2launch.html) sysprep step (responsible for preparing the Windows OS to be reused as a template, resetting the temp dir, temp network interfaces, setting up generic drivers, etc.).

Despite the build failure gone with the January 2023 Windows base templates, it was the 3rd time in 3 months this issue happened (3 failing templates for 6 published, 2 per month).

So this PR is aimed at using the new [EC2 Launch v2](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2launch-v2.html) tools which is supported by both 2019 and 2022 Windows editions.

Please note that it requires a change of the base Windows image: from "Core Containers" to "Full Containers". It adds ~320Mb (compressed) to the final image, with an overhead of roughly 10 more second to start the instance compensated by a faster service startup.


dduportal:
Currently playing around with https://developer.hashicorp.com/packer/docs/debugging locally

dduportal:
Interesting documentation: https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/Creating_EBSbacked_WinAMI.html#sysprep-gui-procedure-ec2launchv2

lemeurherve (approved):

dduportal:
Only the (usual) packer azure plugin crash: the other provisioning are 👍

Issue #337, opened 2022-09-14
(windows): make install not exiting when fail


smerle33:
as per https://infra.ci.jenkins.io/blue/organizations/jenkins/infra-tools%2Fpacker-images/detail/0.41.1/1/pipeline/249#step-306-log-261 the installation of make is not aborting the build when it fail.

smerle33:
this as for consequence to generate an error in the deleting process on azure-arm and windows VM probably due to a delay in stopping the machine.

dduportal:
As per https://github.com/jenkins-infra/packer-images/pull/459#issuecomment-1375865100 and https://ss64.com/ps/call.html, the ampersand (`&`)  character runs the provided instruction as a background job, so the exit code is not captured.

Worth a PR to run the sanity check on a subsequent powershell provider AFTER running the `windows-provision.ps1` provider.

dduportal:
Fixed by https://github.com/jenkins-infra/packer-images/pull/2811/changes#diff-40afc0140eabc567a7fffe74f8e456d714ec88e36e9aca911d2e95cd8696a40bR320

PR #459, opened 2023-01-08
feat: add tfsec (v1.28.1) on all platforms

Kevin-CB:
For https://github.com/jenkins-infra/packer-images/issues/198

This PR add tfsec on all platforms

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -536,7 +536,7 @@ function install_tfsec() {
   apt-get update --quiet
   apt-get install --yes --no-install-recommends curl # Should already be there but this function should be autonomous
 
-  curl --silent --location --show-error "https://github.com/aquasecurity/tfsec/releases/download/v${TFSEC_VERSION}/tfsec-checkgen-linux-${ARCHITECTURE}" --output /usr/local/bin/tfsec
+  curl --silent --location --show-error "https://github.com/aquasecurity/tfsec/releases/download/v${TFSEC_VERSION}/tfsec-linux-${ARCHITECTURE}" --output /usr/local/bin/tfsec
   chmod +rx /usr/local/bin/tfsec
 }
 
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -532,6 +532,14 @@ function install_goss() {
   chmod +rx /usr/local/bin/goss
 }
 
+function install_tfsec() {
+  apt-get update --quiet
+  apt-get install --yes --no-install-recommends curl # Should already be there but this function should be autonomous
+
+  curl --silent --location --show-error "https://github.com/aquasecurity/tfsec/releases/download/v${TFSEC_VERSION}/tfsec-checkgen-linux-${ARCHITECTURE}" --output /usr/local/bin/tfsec
+  chmod +rx /usr/local/bin/tfsec
+}
+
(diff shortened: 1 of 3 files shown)


lemeurherve:
Thanks for this new PR @Kevin-CB 👏 

FYI there are two issues but I haven't found yet why they're occuring:
- the sanity check is apparently running `tfsec-checkgen` instead of `tfsec`, and this tool doesn't know the `--version` flag, thus failing the (Linux) builds
- This sanity check returning the same error message on Windows surprisingly doesn't fail the corresponding Windows builds

I'll take another look later.

provisioning/ubuntu-provision.sh
```diff
@@ -532,6 +532,14 @@ function install_goss() {
   chmod +rx /usr/local/bin/goss
 }
 
+function install_tfsec() {
+  apt-get update --quiet
+  apt-get install --yes --no-install-recommends curl # Should already be there but this function should be autonomous
+
+  curl --silent --location --show-error "https://github.com/aquasecurity/tfsec/releases/download/v${TFSEC_VERSION}/tfsec-checkgen-linux-${ARCHITECTURE}" --output /usr/local/bin/tfsec
```
lemeurherve, line 539:
```suggestion
  curl --silent --location --show-error "https://github.com/aquasecurity/tfsec/releases/download/v${TFSEC_VERSION}/tfsec-linux-${ARCHITECTURE}" --output /usr/local/bin/tfsec
```

I think the fact only tfsec-checkgen release assets are displayed by default induced you in error:
image

provisioning/windows-provision.ps1
```diff
@@ -296,6 +296,13 @@ $downloads = [ordered]@{
             & kubectl.exe version --client;
         }
     };
+    'tfsec' = @{
+        'url' = 'https://github.com/aquasecurity/tfsec/releases/download/v{0}/tfsec-checkgen-windows-amd64.exe' -f $env:TFSEC_VERSION;
```
lemeurherve, line 300:
```suggestion
        'url' = 'https://github.com/aquasecurity/tfsec/releases/download/v{0}/tfsec-windows-amd64.exe' -f $env:TFSEC_VERSION;

```

lemeurherve (commented):
Found the culprit for the first issue.

provisioning/ubuntu-provision.sh
Kevin-CB: Yeah indeed, you're right 😅 

Sorry about that and thanks for the fix!

dduportal:
>     * This sanity check returning the same error message on Windows surprisingly doesn't fail the corresponding Windows builds (FYI @dduportal 👀)

As far as I can tell (ref. https://github.com/jenkins-infra/packer-images/pull/459/files#diff-40afc0140eabc567a7fffe74f8e456d714ec88e36e9aca911d2e95cd8696a40bR2), the powershell script fails when its instructions fails. Just tested it and it works as expected.

I wonder if the mechanism we use for the sanity check is not the culprit: my powershell-fu is not really good but from https://ss64.com/ps/call.html it looks to me that the exit code is not captured.

https://github.com/jenkins-infra/packer-images/issues/337 is the same (I'm adding more note on the issue to avoid polluting the discussion here)

lemeurherve (approved):

dduportal (approved):
Nice job, thanks a lot! 👏

PR #299, opened 2022-07-27
breaking: bump from Ubuntu 20.04 to Ubuntu 22.04

lemeurherve:
Fixes https://github.com/jenkins-infra/packer-images/issues/300

Created the following VM image definitions in portal.azure.com:

<details>
<summary>Details</summary>

image

image

image

</details>

Note: some of the VM image definitions are located in `eastus2` instead of `eastus`.

PR #563, opened 2023-03-27
fix(ubuntu) ensure playwright dependencies are installed by playwright

dduportal:
While working on #299 , it appears that the list of packages changes over OS versions and playwright versions.

This PR ensured that we install the dependencies specified by the latest playwright version.

Note that it requires NPM to be up to date.


dduportal:
Ping @timja , given playwright looks able to manage Chromium and other web browsers: is this still needed to install Chromium through packages ?

PR #564, opened 2023-03-27
chore(ubuntu) use local gpg public keys instead of downloading

dduportal:
While working on https://github.com/jenkins-infra/packer-images/pull/299 , it appears that:

- Hashicorp tools installation does not follow modern Ubuntu GPG armoring (`apt-key` is deprecated)
- Both datadog-agent and Hashicorp are downloading GPG public keys on each build instead of using a "controlled" one

Please note that this is also an excuse to cleanup some minor (but annoying) errors around packages installation for gnupg


provisioning/ubuntu-provision.sh
```diff
@@ -679,7 +677,7 @@ function main() {
   install_JA_requirements
   install_qemu
   install_python
-  install_git_gitlfs
+
```
lemeurherve, line 680:
```suggestion
```

lemeurherve (approved):

Issue #570, opened 2023-03-31
Windows + Azure: Docker installation fails


dduportal:
Since the release of Docker 23.0.2, the main branch build fails on the Windows images on Azure  (both 2019 and 2020).

The error is the following:

```
[2023-03-29T09:21:58.863Z] 2023-03-29T09:21:58Z:     azure-arm.windows: == Setting up Docker Package...
[2023-03-29T09:22:20.973Z] 2023-03-29T09:22:18Z:     azure-arm.windows: WARNING: Cannot find path 'C:\Users\packer\AppData\Local\Temp\DockerMsftProvider\DockerDefault_DockerSearchIndex.json'
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: because it does not exist.
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: WARNING: The remote server returned an error: (404) Not Found.
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: Install-Package : No match was found for the specified search criteria and package name 'docker'. Try
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: Get-PackageSource to see all available registered package sources.
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: At C:\Windows\Temp\script-64240305-9cdc-4be3-ead8-6f5bf6fa811b.ps1:89 char:5
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: +     Install-Package -Name docker -ProviderName DockerMsftProvider -Fo ...
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows: +     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows:     + CategoryInfo          : ObjectNotFound: (Microsoft.Power....InstallPackage:InstallPackage) [Install-Package], Ex
[2023-03-29T09:22:20.974Z] 2023-03-29T09:22:18Z:     azure-arm.windows:    ception
[...]

PR #571, opened 2023-03-31
feat(windows) ensure Docker-CE installation is pinned

dduportal:
This PR comes from #570.

It was changed from a "bugfix" to "enhancement".

It ensures that all images install a pinned version of Docker-CE.

On Azure Windows templates, it switches from the Mirantis Docker runtime to the Community Edition.


smerle33 (approved):
🚀

PR #587, opened 2023-04-11
chore(cleanup AWS) garbage collect EC2 snapshots older than 1 year on `us-east-2`

dduportal:
Implements https://github.com/jenkins-infra/helpdesk/issues/2846#issuecomment-1120206965, in the context of https://github.com/jenkins-infra/helpdesk/issues/3502.

This PR add a GC for AWS snapshots, restricted to `us-east-2` .

### DIFF
--- a/cleanup/aws_images.sh
+++ b/cleanup/aws_images.sh
@@ -41,7 +41,7 @@ aws sts get-caller-identity >/dev/null || \
 
 ## STEP 1
 ## Remove images older than 1 month from dev
-INSTANCE_IDS="$(aws ec2 describe-images --owners self  --filters 'Name=tag:build_type,Values=dev' \
+INSTANCE_IDS="$(aws ec2 describe-images --owners self --filters 'Name=tag:build_type,Values=dev' \
   --query 'Images[?CreationDate<=`'"${lastmonthdate}"'`][].ImageId' | jq -r '.[]' | xargs)"
 
 if [ -n "${INSTANCE_IDS}" ]
--- a/cleanup/aws_images.sh
+++ b/cleanup/aws_images.sh
@@ -40,7 +40,7 @@ aws sts get-caller-identity >/dev/null || \
   { echo "[ERROR] Unable to request the AWS API: the command 'sts get-caller-identity' failed. Please check your AWS credentials"; exit 1; }
 
 ## STEP 1
-## Remove images older than 1 month from dev
+## Remove images older than <timeshift_month> month(s) from dev
 INSTANCE_IDS="$(aws ec2 describe-images --owners self --filters 'Name=tag:build_type,Values=dev' \
   --query 'Images[?CreationDate<=`'"${lastmonthdate}"'`][].ImageId' | jq -r '.[]' | xargs)"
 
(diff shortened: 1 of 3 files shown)


dduportal:
Initial build was too long (more than 20 min and counting). Let's start slowly with a 12month threshold for the first cleanup operation.

dduportal:
- With the "1 year" threshold, the dry run GC for snapshots took ~12 min. Let's start with this value.
- Stopped the build (as no need to consume VMs for this script to run).

cleanup/aws_snapshots.sh
```diff
@@ -0,0 +1,71 @@
...
+    # GNU systems (Linux)
+    start_time_threshold="$(date --date="-${timeshift_month} months" +%Y-%m-%d)"
+fi
+
+## Check for aws API reachibility (is it configured?)
+aws sts get-caller-identity >/dev/null || \
+  { echo "[ERROR] Unable to request the AWS API: the command 'sts get-caller-identity' failed. Please check your AWS credentials"; exit 1; }
+
+## STEP 1
+## Remove snapshots older than 1 month from dev
```
lemeurherve, line 42:
```suggestion
## Remove snapshots older than <timeshift_month> month(s) from dev
```

lemeurherve (commented):
LGTM, wondering if regrouping the two scripts together would worth it 🤔 
(It seems to me they are almost similar except for the query, the cleanup command parameter and the subject of the log messages)

cleanup/aws_images.sh
```diff
@@ -41,7 +41,7 @@ aws sts get-caller-identity >/dev/null || \
 
 ## STEP 1
 ## Remove images older than 1 month from dev
-INSTANCE_IDS="$(aws ec2 describe-images --owners self  --filters 'Name=tag:build_type,Values=dev' \
+INSTANCE_IDS="$(aws ec2 describe-images --owners self --filters 'Name=tag:build_type,Values=dev' \
```
lemeurherve, line 44: How about the other build_type than `dev`?

cleanup/aws_snapshots.sh
```diff
@@ -0,0 +1,71 @@
...
+    # BSD systems (Mac OS X)
+    start_time_threshold="$(date -v-${timeshift_month}m +%Y-%m-%d)"
+else
+    # GNU systems (Linux)
+    start_time_threshold="$(date --date="-${timeshift_month} months" +%Y-%m-%d)"
+fi
+
+## Check for aws API reachibility (is it configured?)
+aws sts get-caller-identity >/dev/null || \
+  { echo "[ERROR] Unable to request the AWS API: the command 'sts get-caller-identity' failed. Please check your AWS credentials"; exit 1; }
```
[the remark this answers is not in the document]
dduportal: Would risk leaking credentials informations. It's already really advanced to point this kind of error to end user.

cleanup/aws_images.sh
lemeurherve: I see why dev, staging and prod cases are separated, thanks for the link.

dduportal:
> LGTM, wondering if regrouping the two scripts together would worth it 🤔 (It seems to me they are almost similar except for the query, the cleanup command parameter and the subject of the log messages)

The goal is to keep a list of simpler scripts to be run independantly of each other.
There are also slight difference between commands: we could factorize but it's not the effort on my opinion

PR #589, opened 2023-04-11
Chore/fixup cleanup aws snapshots

dduportal:
Fixup of https://github.com/jenkins-infra/packer-images/pull/587

- Switches to the "all in one" container image to allow using `parallel` CLI
  - Shell tools such as `jq`, `xargs` are present
  - Cloud CLIs for AWS (`aws`)  and Azure (`az`)  are present and kept up to date

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -54,7 +54,7 @@ spec:
   automountServiceAccountToken: false
   containers:
     - name: jnlp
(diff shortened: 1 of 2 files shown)


dduportal:
- ~9min with `parallel` instead of ~11min without: a bit of improvement
- No error, and `parallel` is explictly told to not fail fast when having an error. Tried locally with no credentials disabled while running: errors are reported but it seems to continue doing its job.

PR #591, opened 2023-04-11
Bump terraform to `1.4.4`

timja:
https://github.com/jenkins-infra/azure/pull/323/commits/0089731d3ee0ae793dbe58fdd7e4297b1d557771 failed because https://developer.hashicorp.com/terraform/language/functions/endswith was introduced in 1.3.x which I would've expected to be available.


dduportal:
We did not switched (yet) the Terraform build agents to the "all in one" image (ref. https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L12). 
Rationale is because the image here misses at least Golang (for the terratest smoke tests) - https://github.com/jenkins-infra/packer-images/issues/198.


For now, upgrading Terraform is tracked in https://github.com/jenkins-infra/docker-hashicorp-tools/pull/193 *but*:
- It was for 1.1.x -> 1.3.x. Might be directly to 1.4.x line.
- We never took the time to perform the upgrade: main reason is because it requires upgrading all Terraform projects one by one (which is necessary to avoid breaking things like we did with the 1.0 -> 1.1...)

dduportal (approved):

PR #594, opened 2023-04-12
chore(cleanup AWS images) remove AMI with build_type=prod older than 2 months

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/2846


dduportal:
- Build stopped after the GC
- GC on AWS seems to work : a lof of AMI were picked up for deletiong (dry run) in the `prod` stage.
Let's go

PR #597, opened 2023-04-12
chore(cleanup AWS snapshots) try to delete when older than 1 day

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2846 and https://github.com/jenkins-infra/helpdesk/issues/3502

Since the AMI for `build_type=dev` are cleaned up after 24h, we have to do the same for snapshots.

Some figures: With the 2 mointh threshold for snapshots, we still have ~1000 snapshots (that's a LOT of packer builds :D )

### DIFF
--- a/cleanup/aws_snapshots.sh
+++ b/cleanup/aws_snapshots.sh
@@ -25,13 +25,13 @@ do
 done
 
 start_time_threshold=""
-timeshift_month="2"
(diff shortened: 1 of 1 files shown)


dduportal:
Please note that this PR might lead the step to failures (as the captured snapshots might be associated to existing AMIs)

PR #598, opened 2023-04-12
chore: ensure that spot instances are used for packer builds

dduportal:
As per https://github.com/jenkins-infra/helpdesk/issues/3502 and https://github.com/jenkins-infra/helpdesk/issues/3485, cloud cost control is important for us.

The goal of this PR is to ensure that packer VM used to build VM templates are always using Spot prices on both AWS EC2 and Azure.

### DIFF
--- a/build-jenkins-agent-ubuntu.pkr.hcl
+++ b/build-jenkins-agent-ubuntu.pkr.hcl
@@ -16,7 +16,6 @@ build {
     # List available SKUs with the command `az vm image list-skus --offer 0001-com-ubuntu-server-jammy --location eastus --publisher canonical --output table`
     image_sku = "${local.agent_os_version_safe}-lts-gen2"
     os_type   = "Linux"
-    vm_size   = local.azure_vm_size
   }
 
   provisioner "shell" {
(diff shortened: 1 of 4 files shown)


timja (approved):

smerle33 (approved):
LGTM

smerle33 (changes requested):
there is an error with the spot

dduportal:
Weird, it points to the `spot = true` in the `azure-arm`  source, which I got from https://developer.hashicorp.com/packer/plugins/builders/azure/arm#spot .

I might be mis-reading the doc, lemme try locally

dduportal:
Oh I see, there documentation is incomplete: https://github.com/hashicorp/packer-plugin-azure/blob/79cd1eab80ead3910cf7507ef8355766c816818d/builder/azure/arm/config_test.go#L2644-L2646

dduportal:
Opened a PR on their doc: https://github.com/hashicorp/packer-plugin-azure/pull/293

dduportal:
- AWS spot problem reproduced locally 
- We are hitting https://github.com/hashicorp/packer-plugin-amazon/issues/198 
- Adding the `spot_price` atrribute explictly set to `auto` allows me to spin up an EC2 `arm64` VM, so let's try this on the CI now

PR #599, opened 2023-04-13
fix(windows) ensure NugetPackage Provider is present

dduportal:
Related to https://github.com/jenkinsci/docker-agent/pull/399

This PR fixes the issue by ensuring Nuget is properly installed in *non-interactive*.
It makes this dependency explicit.

This was caused by #571 which removed the Docker installation through the `Msft` provider which was installing Nuget as a dependency.
The new Docker CE installation method does not: that's why `Nuget` is absent since the `1.1.0` image.


timja (approved):
nice find

PR #600, opened 2023-04-14
feat(azure/arm): enable arm64 azure build

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3471 and following #575

enable ubuntu 22.04 ARM on azure but exclude windows build in arm

PR #601, opened 2023-04-14
fix(imagename): add the cpu architecture to the name

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3471
adding the architecture to the managed image name should solved the arm/amd problem
and adding the timestamp should help on the concurrent access on image with the same name during development


smerle33:
superseded by #600

PR #605, opened 2023-04-17
feat: add Launchable

lemeurherve:


provisioning/ubuntu-provision.sh
```diff
@@ -567,6 +567,14 @@ function install_playwright_dependencies() {
   rm -rf "${temp_dir}"
 }
 
+# Install Launchable with python3
+function install_launchable() {
+  # Install Launchable with python3 in its own virtual environment
+  python3 -m venv launchable
+  launchable/bin/pip --require-virtualenv --no-cache-dir install setuptools wheel
+  launchable/bin/pip --require-virtualenv --no-cache-dir install launchable==${LAUNCHABLE_VERSION}
```
dduportal, line 575:
```suggestion
  # Install Launchable with python3 in its own virtual environment
  local venv_path=/launchable
  python3 -m venv "${venv_path}"
  "${venv_path}"/bin/pip --require-virtualenv --no-cache-dir install setuptools wheel
  "${venv_path}"/bin/pip --require-virtualenv --no-cache-dir install launchable=="${LAUNCHABLE_VERSION}"
```

WDYT about using a variable to avoid repetition and make it clearer?

Also:
- Use absolute path (recommended to make the function invulnerable to working dir change)
- add double quotes around the `LAUNCHABLE_VERSION` variable (as its content come from the external environment)

lemeurherve:
Sanity checks OK on Windows and Ubuntu ✅

provisioning/ubuntu-provision.sh
```diff
@@ -568,12 +568,12 @@ function install_playwright_dependencies() {
...
 }
 
-# Install Launchable with python3
+## Install Launchable with python3 in its own virtual environment
 function install_launchable() {
-  # Install Launchable with python3 in its own virtual environment
   python3 -m venv "${launchable_venv_dir}"
   "${launchable_venv_dir}"/bin/pip --require-virtualenv --no-cache-dir install setuptools wheel
   "${launchable_venv_dir}"/bin/pip --require-virtualenv --no-cache-dir install launchable=="${LAUNCHABLE_VERSION}"
+  # Simlink to common tools folder
```
dduportal, line 576: nit:

```suggestion
  # symlink to a folder which is in the PATH
```

or

```suggestion
  # Symbolic link to a folder which is in the PATH
```

provisioning/ubuntu-provision.sh
```diff
@@ -567,6 +568,15 @@ function install_playwright_dependencies() {
   rm -rf "${temp_dir}"
 }
 
+## Install Launchable with python3 in its own virtual environment
+function install_launchable() {
+  python3 -m venv "${launchable_venv_dir}"
+  "${launchable_venv_dir}"/bin/pip --require-virtualenv --no-cache-dir install setuptools wheel
+  "${launchable_venv_dir}"/bin/pip --require-virtualenv --no-cache-dir install launchable=="${LAUNCHABLE_VERSION}"
+  # Simlink to a folder in the PATH
```
dduportal, line 576:
```suggestion
  # Symlink to a folder in the PATH
```

dduportal (approved):

PR #606, opened 2023-04-18
chore: separated updatecli pipeline

lemeurherve:
Needs https://github.com/jenkins-infra/kubernetes-management/pull/3855

Ref: https://github.com/jenkins-infra/helpdesk/issues/2778


lemeurherve:
Corresponding builds: https://infra.ci.jenkins.io/job/updatecli/job/packer-images/view/change-requests/job/PR-606/

dduportal (changes requested):
That is a good idea, but you can't merge "as it": this PR break the whole updatecli for this repository, because it changes the environment in which `updatecli` was executed before so it could, or could not, break things.

The current `main` branch currently executes `updatecli` inside a VM agent which label is `linux-amd64-docker`:

- Top-level agent directive (e.g. default agent): https://github.com/jenkins-infra/packer-images/blob/302702dcca4d178c6bd01b9750a522465d36367d/Jenkinsfile_k8s#L64
- No `agent` directive in the `Side Task` "top level" stage: https://github.com/jenkins-infra/packer-images/blob/302702dcca4d178c6bd01b9750a522465d36367d/Jenkinsfile_k8s#L75-L149
- No `agent` directive in the `updatecli` stage: https://github.com/jenkins-infra/packer-images/blob/302702dcca4d178c6bd01b9750a522465d36367d/Jenkinsfile_k8s#L126-L145

Also, don't forget that the `updatecli()` pipeline library method takes care of spawning its own pod agent (default image is https://github.com/jenkins-infra/pipeline-library/blob/9e0f0a2f09a7c91b26824e706a8963598b685742/vars/updatecli.groovy#L10, you can change it)

=> There was a good reason to NOT use the pipeline library before (the comment points to an issue :) )
=> Proposal: use the same pipeline code as before , but in the new pipeline for this PR to work.
And/or implement the feature request in the shared library.

=> both case needs you to use `agent { label ... }` to spin up a VM with Docker inside otherwise you break https://github.com/jenkins-infra/packer-images/blob/main/updatecli/updatecli.d/docker-ce.yaml

Jenkinsfile_updatecli
```diff
@@ -2,11 +2,8 @@ def cronExpr = env.BRANCH_IS_PRIMARY ? 'H/30 * * * *' : ''
 
 pipeline {
   agent {
-    kubernetes {
-      yamlFile 'podTemplate.yaml'
-    }
+    docker { image 'jenkinsciinfra/helmfile:2.5.92' }
```
dduportal, line 5:
```suggestion
    label "linux-amd64-docker"
```

This updatecli process requires a Docker Engine, which means it requires a VM.

PR #635, opened 2023-04-27
chore(azure cleanup) split gallery image and instance cleanup + use time threshold for gallery images

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3552

This PR is a preliminary work for the fix of Azure resources cleanup for single resource group (caused by https://github.com/jenkins-infra/packer-images/pull/552).

The monolitic  script `cleanup/azure.sh` is a nightmare to use: no option to select only one of its sections, hard to read, etc.
Also, the gallery versions cleanup does not use the timestamp threshold: it deletes "all".


dduportal:
- Aborted the build as we only need to verify the GC
- The GC for azure did not fail and prints "no dangling" images as expected

lemeurherve:
> Also, the gallery versions cleanup does not use the timestamp threshold: it deletes "all".

This is intended, no need for these staging/dev gallery image versions.

lemeurherve:
I would not have changed it: the cleanup already occurs only one time per day, there is already enough time to test these in that timeframe.

lemeurherve:
More simple without it, but fine if added.

dduportal:
> I would not have changed it: the cleanup already occurs only one time per day, there is already enough time to test these in that timeframe.

The cleanup happens by default once a day.
But it also happens on eahc build of the main branch (e.g. each time a PR is merged).

I was bitten by this behavior yesterday as I searched for the image to diagnose an issue but it was not quick enough and the image was gone.

PR #636, opened 2023-04-27
chore(azure cleanup) remove resources from the packer-builds RGs

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3552, this PR cleans up ALL resources older than the provided threshold in days for each of the 3 RGs (including VM, vaults, disks, public IP, NIC, etc.)

Please note I chose to call the script 3 times (each RG) with the same threshold. The goal is to surface, at pipeline level, the parameters (instead of hiding it inside the script). 
It has the benefit to make the script simpler, less loops, so the execution time and logs are easier to gather.


dduportal:
- Tested locally in dry run
- Ran manually one of the "dry runned" command with success

lemeurherve (approved):

PR #664, opened 2023-05-22
fixup(pipeline): remove pod yaml in favor of default arm64 agent

smerle33:
in order to solve the mismatch ARM/AMD on GC


dduportal (approved):
LGTM!

That will solve in short term the `exec /usr/local/bin/jenkins-agent: exec format error` error we see when a given release of packer-image does not generate the Docker manifest at the end, which is an "egg and chicken" class of problem.

PR #701, opened 2023-06-02
chore(updatecli): remove double quotes to avoid PR storm from new yaml parser

smerle33:
ref https://www.bram.us/2022/01/11/yaml-the-norway-problem/


smerle33:
tested locally

PR #873, opened 2023-10-24
tests(windows) fix JDK8 goss check

dduportal:


dduportal:
Self merging: it is an hotfix

PR #835, opened 2023-10-10
cleanup(jdk19): remove jdk19

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3289
jdk19 is EOL

PR #897, opened 2023-11-13
chore(tests) fix goss linux harness

dduportal:
This PR is a quick fixup of #835 

- (nit) sort execution targets
- (fix) call `node --version` for the NodeJS version check
- (fix) remove stdout assertions for tools which version is not pinned
- (feat) add a check for the default Java version

### DIFF
--- a/goss/goss-linux.yaml
+++ b/goss/goss-linux.yaml
@@ -1,4 +1,61 @@
 command:
+  asdf:
+    exec: asdf version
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

dduportal:
Main checks succeeded, let's roll (updatecli will be fixed in a subsequent PR)

PR #898, opened 2023-11-13
feat: add `azcopy` to Linux provisioning

lemeurherve:
This PR adds https://github.com/Azure/azure-storage-azcopy to Linux provisioning so we can use it in https://github.com/jenkins-infra/crawler/ job to synchronize the Update Center File Share that will be used by updates.jenkins.io.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-1807816044

### DIFF
--- a/goss/goss-linux.yaml
+++ b/goss/goss-linux.yaml
@@ -52,6 +52,11 @@ command:
     exit-status: 0
     stdout:
       - 2.53.1
+  azcopy:
+    exec: azcopy --version
+    exit-status: 0
+    stdout:
+      - 10.21.0
   bundle:  # not updatecli tracked
     exec: bundle -v
     exit-status: 0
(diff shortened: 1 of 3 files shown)


goss/goss-linux.yaml
```diff
@@ -14,6 +14,11 @@ command:
     exit-status: 0
     stdout:
       - 2.53.1
+  azcopy:
```
lemeurherve, line 17:
```suggestion
  azcopy: # not tracked by updatecli
```

provisioning/ubuntu-provision.sh
```diff
@@ -403,6 +403,15 @@ function install_azurecli() {
   apt-get install --yes --no-install-recommends azure-cli="${AZURECLI_VERSION}*"
 }
 
+## Ensure that azcopy is installed
+function install_azcopy() {
```
dduportal, line 407: Don't forget to call the function 😅

dduportal (changes requested):
Almost there: you have to call the function

provisioning/ubuntu-provision.sh
dduportal: Fixup commit pushed (https://github.com/jenkins-infra/packer-images/pull/898/commits/3f832ab738b637989ab74b5648af61362dc56ed2)

dduportal (approved):

PR #906, opened 2023-11-14
chore(goss): moving all sanity checks to goss

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3763#issuecomment-1812019058
phase 2 of migration within goss

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -615,63 +615,6 @@ function cleanup() {
   sync
 }
 
-function sanity_check() {
-  echo "== Sanity Check of installed tools, running as user ${username}"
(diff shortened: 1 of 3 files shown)


provisioning/ubuntu-provision.sh
```diff
@@ -615,63 +615,6 @@ function cleanup() {
...
-  && echo 'npm version:' \
-  && npm --version \
-  && echo 'playwright version:' \
-  && playwright --version \
-  && echo 'launchable version:' \
-  && launchable --version
-  "
-  echo "== End of sanity check"
-  echo "== Installed packages:"
-  dpkg -l
```
dduportal, line 672: Do you mind keeping this instruction : it is not a sanity check per-se, but it helps for auditing software versions for packages (proposal : run it after calling the function `main()` below)

PR #910, opened 2023-11-15
chore(updatecli): update to keep goss file up2date

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3763

following https://github.com/jenkins-infra/packer-images/pull/906
update the corresponding updatecli manifest to keep the goss file on track.


dduportal (approved):
LGTM. The 2 failures looks like an edge case of `updatecli` as we are not able to reproduce on our local setups. Let's merge and check if the problem is gone or not in the principal branch

PR #919, opened 2023-11-21
chore(updatecli, tests) fix Linux Git manifest to also update goss

dduportal:
- Updates the version of git in the Linux Goss test harness (fixup of #893)
- Fixes the `updatecli` manifest to update Git version in the Linux Goss test harness (fixup of #910)


MarkEWaite (approved):
Thanks!

PR #949, opened 2023-12-18
feat(doctl): add doctl install on linux only

smerle33:
as per https://github.com/jenkins-infra/packer-images/issues/198

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -348,6 +348,14 @@ function install_docker_compose(){
   chmod a+x /usr/local/bin/docker-compose
 }
 
+## Ensure that DOCTL is installed
+function install_doctl(){
+  curl --fail --silent --location --show-error --output /usr/local/bin/doctl \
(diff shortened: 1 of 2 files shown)


provisioning/ubuntu-provision.sh
```diff
@@ -348,6 +348,14 @@ function install_docker_compose(){
   chmod a+x /usr/local/bin/docker-compose
 }
 
+## Ensure that DOCTL is installed
+function install_doctl(){
+  curl --fail --silent --location --show-error --output /usr/local/bin/doctl \
+    "https://github.com/digitalocean/doctl/releases/download/v${DOCTL_VERSION}/doctl-${DOCTL_VERSION}-linux-$(uname -m).tar.gz" | tar -xzv doctl
```
dduportal, line 354: This command is invalid:

- `curl` downloads the tar gz archive to `/usr/local/bin/doctl`
- There is a `tar` command piped from the `curl` which does nothing (as there are no output on the stdout from `curl`)

dduportal (changes requested):
Installation command cannot work as it, you might want to try it inside an `ubuntu:22.04` container

dduportal (approved):

dduportal (approved):
Merging as the only error is due to #944 and was fixed by #952

PR #981, opened 2024-01-11
feat: add golang and golangcilint on linux agent

smerle33:
as per https://github.com/jenkins-infra/packer-images/issues/198

install golang and golandci-lint on the allinone agent.


provisioning/ubuntu-provision.sh
```diff
@@ -198,6 +198,23 @@ function install_docker() {
...
 
+## Ensure Golang is installed
+function install_golang(){
+  local temp_archive="/tmp/go${GOLANG_VERSION}.tgz"
+  local installation_dir="/usr/local/"
+  local download_url="https://go.dev/dl/go${GOLANG_VERSION}.linux-${ARCHITECTURE}.tar.gz"
+  curl --fail --silent --show-error --location --output "${temp_archive}" "${download_url}"
+  tar --extract --gunzip --file="${temp_archive}" --directory="${installation_dir}"
+  # Symlink to a folder in the PATH
+  ln -s "${installation_dir}go/bin/go" /usr/local/bin/go
```
lemeurherve, line 209: Are there other files in the downloaded archive? If not or if these files aren't needed, I would extract only the go binary, directly to /usr/local/bin/go so the symbolic link wouldn't be needed anymore.

provisioning/ubuntu-provision.sh
```diff
@@ -198,6 +198,23 @@ function install_docker() {
...
+}
+
+## Ensure GolangCIlint is installed
+function install_golangcilint(){
+  # binary will be in /usr/local/bin
+  curl -sSfL "https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh" | sh -s -- -b /usr/local/bin "v${GOLANGCILINT_VERSION}"
```
lemeurherve, line 215: Not sure about the execution of this external script.

lemeurherve (commented):
Some comments and questions.

provisioning/ubuntu-provision.sh
lemeurherve: I would extract the main commands/logic from this remote script to our own script here. (download from a versioned URL + checksum + tar)

dduportal (approved):

PR #988, opened 2024-01-15
chore(goss): split linux and windows common goss tests to a common file

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3763

PR #989, opened 2024-01-15
chore(updatecli): update updatecli manifests to use the new goss-common.yaml file

smerle33:
follow up #988 to update updatecli 

as per https://github.com/jenkins-infra/helpdesk/issues/3763


provisioning/windows-provision.ps1
```diff
@@ -174,7 +174,7 @@ $downloads = [ordered]@{
         'cleanupLocal' = 'true';
     };
     'jq' = @{
-        'url' = 'https://github.com/stedolan/jq/releases/download/jq-{0}/jq-win64.exe'  -f $env:JQ_VERSION;
+        'url' = 'https://github.com/jqlang/jq/releases/download/jq-{0}/jq-win64.exe'  -f $env:JQ_VERSION;
```
lemeurherve, line 177: I would add something about this change in the PR body.

Was it required for this PR or could it have been its own?

lemeurherve (approved):

provisioning/windows-provision.ps1
lemeurherve: Thanks for the explanation. I still would add a mention of this change in the PR body 😉

lemeurherve (approved):

PR #1004, opened 2024-01-18
hotfix(docker): add golang to path

smerle33:
follow #981 as per https://github.com/jenkins-infra/helpdesk/issues/3823#issuecomment-1887188512


dduportal (approved):

smerle33:
force merged as the error is only on windows

PR #1017, opened 2024-01-24
chore: regroup consecutive `sh` calls

lemeurherve:
This PR regroups consecutive `sh` calls.


dduportal (changes requested):
There is a reason for splitting the call: make it easier to measure time spent on each step.

lemeurherve:
TIL

This isn't obvious, maybe worthy of a comment in the pipeline header?

lemeurherve:
> make it easier to measure time spent on each step.

I don't know how much it makes a difference between consecutive sh calls versus one sh call, but if the purpose is to measure time spent, how about prefixing important commands with `time ` instead?

PR #1021, opened 2024-01-26
feat(node): change npm version check and add asdf to path

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3823#issuecomment-1908570581

correct path in order to use nodejs/npm or bundle as in https://github.com/jenkins-infra/infra-reports/blob/3076337651c78c2ecc40155f3856534fbfbbc5b5/fork-report/Jenkinsfile#L28:

### DIFF
--- a/goss/goss-common.yaml
+++ b/goss/goss-common.yaml
@@ -78,10 +78,14 @@ command:
     exec: node --version
     exit-status: 0
     stdout:
-      - 18.18.2
+      and:
+        - 18.18.2
+        - not: 'Please install a version by running one of the following'
   npm:
     exec: npm --version
     exit-status: 0
+    stdout:
+      - not: 'No preset version installed for command npm'
   packer:
     exec: packer -v
     exit-status: 0
--- a/goss/goss-common.yaml
+++ b/goss/goss-common.yaml
@@ -78,14 +78,15 @@ command:
     exec: node --version
     exit-status: 0
     stdout:
-      and:
-        - 18.18.2
-        - not: 'Please install a version by running one of the following'
+      - 18.18.2
+      # and:
+      #   - contain: 18.18.2
+      #   - not: 'Please install a version by running one of the following'
   npm:
     exec: npm --version
     exit-status: 0
-    stdout:
-      - not: 'No preset version installed for command npm'
+    # stdout:
+    #   - not: 'No preset version installed for command npm'
   packer:
     exec: packer -v
     exit-status: 0
--- a/goss/goss-common.yaml
+++ b/goss/goss-common.yaml
@@ -80,7 +80,7 @@ command:
     stdout:
       - 18.18.2
   npm:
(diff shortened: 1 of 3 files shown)


goss/goss-common.yaml
```diff
@@ -79,9 +79,14 @@ command:
     exit-status: 0
     stdout:
       - 18.18.2
+      # and:
```
smerle33, line 82: commented until: https://github.com/goss-org/goss/issues/872

goss/goss-common.yaml
```diff
@@ -79,9 +79,14 @@ command:
...
+      #   - contain: 18.18.2
+      #   - not: 'Please install a version by running one of the following'
   npm:
-    exec: npm version
+    exec: npm --version
     exit-status: 0
+    # stdout:
```
smerle33, line 88: commented until: https://github.com/goss-org/goss/issues/872

provisioning/ubuntu-provision.sh
```diff
@@ -154,6 +154,9 @@ function install_asdf() {
   echo ". ${asdf_install_dir}/asdf.sh" >> "${profile_script}"
   chown -R "${username}:${groupname}" "${userhome}"
   rm -f "${archive}"
+
+  ## append to the system wide path variable, need to be seconded for docker in packer sources.pkr.hcl
+  sed -e "/^PATH/s/\"$/:\${asdf_install_dir}\/shims:\${asdf_install_dir}\/bin\"/g" -i /etc/environment
```
lemeurherve, line 159: There is a `\` in front of the first `$` here while there isn't any line 212, is it expected?

smerle33: it's a double quote interpolated string while the other one is without interpolation, if I surround everything by simple quotes, it will disable interpolation. The `\` is to escape the used double quote `"`

provisioning/ubuntu-provision.sh
```diff
@@ -154,6 +154,9 @@ function install_asdf() {
   echo ". ${asdf_install_dir}/asdf.sh" >> "${profile_script}"
   chown -R "${username}:${groupname}" "${userhome}"
   rm -f "${archive}"
+
+  ## append to the system wide path variable, need to be seconded for docker in packer sources.pkr.hcl
```
lemeurherve, line 158: This comment is the same as the one L211, is it expected?

lemeurherve (approved):

provisioning/ubuntu-provision.sh
smerle33: yes

lemeurherve:
Seems to have caused https://github.com/jenkins-infra/packer-images/issues/1042

dduportal:
> Seems to have caused #1042

I disagree. Remains to be demonstrated.

PR #1043, opened 2024-02-10
fix(ubuntu) correct asdf inclusion in `/etc/environment` for Linux VMs

lemeurherve:
This PR uses hardcoded path to include asdf in the $PATH so asdf installed tools like yq can be executed.

Note: there might be a way to keep this folder as a variable but as it breaks current uses this "easy" solution should be enough, can be revised in a following pull request.

Ref:
- #1042

Issue #1042, opened 2024-02-10
`asdf` isn't in the $PATH, none of asdf installed tools are executable


lemeurherve:
### Reproduction steps

Run this pipeline on infra.ci.jenkins.io:

```groovy
pipeline {
    agent { label 'linux-amd64-docker' }
    stages {
        stage('current') {
            steps {
                sh 'echo "PATH: ${PATH}"'
                sh 'asdf --version || echo "no asdf"'
                sh 'yq --version || echo "no yq"'
            }
        }
        stage('fixed PATH') {
            environment {
                PATH = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/jenkins/.asdf/shims:/home/jenkins/.asdf/bin:/usr/local/go/bin"
            }
            steps {
                sh 'asdf --version || echo "no asdf"'
                sh 'yq --version || echo "no yq"'
            }
        }
    }    
}
```

### Expected Results

Both stages show versions: correct $PATH value, `asdf` and `yq` directly executable without fixing $PATH

### Actual Results

Only the "fixed PATH" shows versions:

> \+ asdf --version
>  v0.14.0
>  [Pipeline] sh
>  \+ yq --version
>  yq (https://github.com/mikefarah/yq/) version 4.25.3

The "current" one doesn't:

> \+ echo 'PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:${asdf_install_dir}/shims:${asdf_install_dir}/bin:/usr/local/go/bin'
>  PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:${asdf_install_dir}/shims:${asdf_install_dir}/bin:/usr/local/go/bin
>  [Pipeline] sh
>  \+ asdf --version
>  /home/jenkins/workspace/jobs_docker-confluence-data_main@tmp/durable-f9cb9506/script.sh.copy: line 1: asdf: command not found
>  \+ echo 'no asdf'
>  no asdf
>  [Pipeline] sh
>  \+ yq --version
>  /home/jenkins/workspace/jobs_docker-confluence-data_main@tmp/durable-aa430d29/script.sh.copy: line 1: yq: command not found
>  \+ echo 'no yq'

[...]

PR #1049, opened 2024-02-13
chore: restore ruby version goss check for Linux

lemeurherve:
This PR uses `asdf list ruby` to retrieve the version installed by asdf and not the vagrant one, cf https://github.com/jenkins-infra/packer-images/pull/1043#pullrequestreview-1878135249

Note: I wasn't sure if goss would have kept the linux version on its config merge if I kept the less precise ruby check in goss-common.yaml, hence the move of the existing one to goss-windows.yaml.

Follow-up of:
- #1043 

Ref:
- #1042

### DIFF
--- a/goss/goss-windows.yaml
+++ b/goss/goss-windows.yaml
@@ -41,6 +41,8 @@ command:
   ruby:
     exec: ruby -v
     exit-status: 0
+    stdout:
+      - 2.6.10
   vagrant:
     exec: vagrant --version
     exit-status: 0
--- a/goss/goss-windows.yaml
+++ b/goss/goss-windows.yaml
@@ -38,6 +38,9 @@ command:
     exit-status: 0
     stdout:
       - 3.12.2
+  ruby:
+    exec: ruby -v
+    exit-status: 0
   vagrant:
     exec: vagrant --version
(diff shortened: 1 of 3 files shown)


lemeurherve:
> [DEBUG][Command][ruby][stdout]   *2.6.10[0m

dduportal (approved):

goss/goss-windows.yaml
```diff
@@ -38,6 +38,9 @@ command:
     exit-status: 0
     stdout:
       - 3.12.2
+  ruby:
+    exec: ruby -v
+    exit-status: 0
```
lemeurherve, line 43:
```suggestion
    exit-status: 0
   stdout:
     - 2.6.10
```
@dduportal Should it be restored too?

dduportal (commented):
Side note: don't forget to check if there are any updatecli manifest associated (I don't think so but better safe than sorry ;) )

lemeurherve:
> Side note: don't forget to check if there are any updatecli manifest associated (I don't think so but better safe than sorry ;) )

There isn't any ruby manifest. We might want to add one later, with version fixed to 2.6.* for Puppet.

PR #1285, opened 2024-07-15
chore(pipeline): diff only on non main branch for updatecli

smerle33:
PR done in pair with @dduportal to accelerate main branch builds.


dduportal:
Merged as the check for `updatecli` passes

PR #1297, opened 2024-07-20
chore(tests,updatecli) fixup OS specific and cleanups

dduportal:
This PR aims at cleaning up the tests and updatecli to fix the following elements:

- chore: Bypass the `updatecli` YAML parsing error when the value of a YAML key is a Go template (one engine [treat it as a string](https://github.com/jenkins-infra/helpdesk/issues/4090), and the other engine we are using parses the template inside updatecli.
  - I've split the Goss Windows tests harnesses in 3: one for all Windows supported host OS and one per major version.
  - Of course, the long term fixes would be to have Visual Studio either removed or installed on both 2019 and 2020 Windows + having updatecli's YAMl parsers NOT to parse gotemplates.
  - It will unblock the Ruby, Python, NodeJS and Temurin PRs in this repository => it is important
- nit: renaming `./goss/` directory to `./tests/` as we can keep the process (e.g. "testing") while changing the tool
- chore: run the tests for Docker CE (version) and Docker BuildX (binary successful execution) on  all hosts (not only Linux)
- chore: test CST version since the 1.19.0 version fixed the missing `version` subcommand

----

Note: updatecli check will fail due to all-in-one PR (e.g. changing both updatecli manifests and file references). Ran the change manifests with the `target.*.scmid` key commented out with success as a test but we'll check the principal branch


dduportal:
Self merging as the packer checks is green: this will trigger a LOT of PRs now \o/ (hopefully).

PR #1370, opened 2024-08-29
feat!(ubuntu): Removed all definitions of 'default_jdk' from ubuntu image templates

jayfranco999:
BREAKING CHANGE: removing the default_jdk env variable will introduce failure in goss tess harness for windows and might result in build failures.

As per https://github.com/jenkins-infra/helpdesk/issues/4124#issuecomment-2307122480 :

Any definition of a 'default JDK' and 'update-alternatives' on the jenkins-infra/packer-images has been removed from the ubuntu-provisioning script.  The corresponding goss test patch has also been added to the goss test harness to check installation of maven.

### DIFF
--- a/tests/goss-common.yaml
+++ b/tests/goss-common.yaml
@@ -27,11 +27,6 @@ command:
   docker_buildx:
     exec: docker buildx version
     exit-status: 0
-  default_java:
-    exec: java --version
-    exit-status: 0
-    stdout:
-      - 11.0.24+8
   docker_compose:
     exec: docker-compose -v
     exit-status: 0
--- a/tests/goss-common.yaml
+++ b/tests/goss-common.yaml
@@ -27,6 +27,10 @@ command:
   docker_buildx:
     exec: docker buildx version
     exit-status: 0
+  default_java:
+    exec: java --version
+    exit-status:
+      not: 0
   docker_compose:
     exec: docker-compose -v
     exit-status: 0
--- a/tests/goss-common.yaml
+++ b/tests/goss-common.yaml
@@ -27,10 +27,6 @@ command:
   docker_buildx:
     exec: docker buildx version
     exit-status: 0
-  default_java:
-    exec: java --version
-    exit-status:
-      not: 0
   docker_compose:
     exec: docker-compose -v
     exit-status: 0
--- a/tests/goss-linux.yaml
+++ b/tests/goss-linux.yaml
@@ -17,6 +17,9 @@ command:
   datadog-agent:
     exec: datadog-agent version
     exit-status: 0
+  default_java:
+    exec: java --version
+    exit-status: 0
   doctl:
     exec: doctl version
     exit-status: 0
@@ -75,7 +78,7 @@ command:
     stderr:
       - 1.8.0_422
   maven:
-    exec: mvn -v
+    exec: JAVA_HOME=/opt/jdk-21 mvn -v
     exit-status: 0
     stdout:
       - 3.9.9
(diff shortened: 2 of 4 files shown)


tests/goss-common.yaml
```diff
@@ -27,11 +27,6 @@ command:
   docker_buildx:
     exec: docker buildx version
     exit-status: 0
```
jayfranco999, line 29:
```suggestion
    exit-status:
      not: 0
```

tests/goss-linux.yaml
```diff
@@ -17,6 +17,9 @@ command:
   datadog-agent:
     exec: datadog-agent version
     exit-status: 0
+  default_java:
+    exec: java --version
+    exit-status: 0
```
dduportal, line 22: This test is failing (Linux only) with the following error:

```
[stderr]  sh: 1: java: not found
```

=> preventing the `goss-common` test to be run (fail fast)

tests/goss-common.yaml
```diff
@@ -26,12 +26,8 @@ command:
...
   docker_buildx:
     exec: docker buildx version
-    exit-status: 0
-  default_java:
-    exec: java --version
-    exit-status: 0
-    stdout:
-      - 11.0.24+8
+    exit-status:
+      not: 0
```
dduportal, line 30: This test is failing on Windows: the command `docker buildx version` does exit with the exit code 0 (which was the old test case).

I'm not sure why did you change the Docker BuildX test case in this PR? Merge conflict mistake?

dduportal (changes requested):
Builds are failing due to test errors: see my comments

tests/goss-common.yaml
jayfranco999: It was a personal error, my bad. Let me fix it.

jayfranco999, line 30:
```suggestion
```

tests/goss-linux.yaml
jayfranco999, line 22:
```suggestion
    exit-status:
      not: 0
```

tests/goss-common.yaml
```diff
@@ -26,12 +26,8 @@ command:
       - 27.2.0
   docker_buildx:
     exec: docker buildx version
-    exit-status: 0
-  default_java:
-    exec: java --version
-    exit-status: 0
-    stdout:
-      - 11.0.24+8
+    exit-status:
```
jayfranco999, line 29:
```suggestion
    exit-status: 0
```

tests/goss-linux.yaml
```diff
@@ -17,6 +17,10 @@ command:
   datadog-agent:
     exec: datadog-agent version
     exit-status: 0
+  default_java:
+    exec: java --version
+    exit-status:
+      not: 0
```
dduportal, line 23: This will fail as you removed any traces of `java` in the default `PATH`

jayfranco999: So should we remove this condition for the test harness? This was included in the test harness that was successful.

dduportal: - what is the intent of this PR?
- Is this change aligned with the intent of this PR?

jayfranco999: removing default jdk, the condition passes if there isn't a default jdk that's picked for linux builds. 
My assumption is that we kept the condition in to see if our changes worked, but now that it does work I'm not sure if we should include this in the PR. Correct me if I'm wrong.

tests/goss-windows.yaml
```diff
@@ -10,6 +10,9 @@ command:
   docker-buildx:
     exec: docker buildx version
     exit-status: 0
+  default_java:
+    exec: java --version
+    exit-status: 0
```
jayfranco999, line 15:
```suggestion
    exit-status: 0
    stdout:
      - 11.0.24+8
```

dduportal (approved):

Issue #1381, opened 2024-09-05
[updatecli] retrieve Maven version from the ATH `Dockerfile` with a native resource


dduportal:
We have a manifest which role is to track the latest version of Maven used on the ATH (jenkinsci/acceptance-test-harness): https://github.com/jenkins-infra/packer-images/blob/f1d6502d4c4599b84403c85134df38b6b41e44ad/updatecli/updatecli.d/maven.yml#L16-L27

This manifest was written **before** updatecli introduced the support of source of kind `Dockerfile` (ref. https://github.com/updatecli/updatecli/issues/2275) so we used a source of type `file` with a transformer to strip out the unneeded parts (`ARG MAVEN_VERSION=` prefix).

This issue is to switch this source to the new source of type `dockerfile` (see the updatecli docs or the related issue) instead to make the manifest easier to read and maintain.

jayfranco999:
As per https://github.com/jenkins-infra/packer-images/issues/1381#issue-2507739764

We switch to the new source of type dockerfile which is easier to read and maintain. ( Ref - https://github.com/jenkins-infra/packer-images/pull/1384 )

Tested and successful locally.

dduportal:
CLosed by #1384

PR #1384, opened 2024-09-06
chore(updatecli):retrieve Maven version from the source with a native resource(dockerfile)

jayfranco999:
As per https://github.com/jenkins-infra/packer-images/issues/1381#issue-2507739764

We switch to the new source of type dockerfile which is easier to read and maintain.

Tested and successful locally.

Also related ( https://github.com/jenkins-infra/helpdesk/issues/4280 )


smerle33 (approved):
updatecli check passes, let's merge 🚀

PR #1443, opened 2024-10-10
chore(updatecli): add manifest for azure windows images

smerle33:
follow up #1442 to keep track of azure windows images

`az vm image list --location eastus --publisher MicrosoftWindowsServer --offer WindowsServer --sku 2019-datacenter-core-g2 --all --query "[?offer=='WindowsServer'].version" -o tsv | sort -u | tail -n 1`

and

`az vm image list --location eastus --publisher MicrosoftWindowsServer --offer WindowsServer --sku 2022-datacenter-core-g2 --all --query "[?offer=='WindowsServer'].version" -o tsv | sort -u | tail -n 1`

------------------------------------------------------------------------------------------
Those commands need authentication ... sic ...

Converted to draft, until next week, where we will provide the authentication ...
done with https://github.com/jenkins-infra/azure/pull/860

and https://github.com/jenkins-infra/kubernetes-management/pull/5806

PR #1472, opened 2024-10-24
chore(ubuntu): Use pinned aws base AMI IDs from `images-versions.yaml`

jayfranco999:
As per https://github.com/jenkins-infra/helpdesk/issues/4365#issue-2610776824

and https://github.com/jenkins-infra/packer-images/pull/1443

We want to use the aws `ubuntu` pinned AMI Ids in `image-versions.yaml`, hence the `source_ami` is retrieved from there.


dduportal (approved):
LGTM! Checks have started: we can proceed.

(note: I'm changing the title to `chore` before merging)

PR #1514, opened 2024-11-12
chore(updatecli) when a new Ruby version is available, check for its existence in the ASDF Ruby plugin

dduportal:
This PR updates the `updatecli` manifest tracking Ruby version to add a condition checking if the latest found version exists in ASDF.

As seen in #1512 , we had to revert because the "latest" available version of Ruby was not (yet) available in ASDF.

Tested with:

- `3.3.5` as source version: works
- `3.3.6` as source version BEFORE https://github.com/asdf-vm/asdf-ruby/pull/419 was merged (or with an out of date ASDF index): conditions fails, no PR created but `updatecli` process continue working as expected
- `3.3.6` as source AFTER https://github.com/asdf-vm/asdf-ruby/pull/419  was merged: `updatecli` wants to create a new PR
- `3.3.7` (not existing when writing this PR): condition fail as expected

### DIFF
--- a/updatecli/updatecli.d/ruby.yml
+++ b/updatecli/updatecli.d/ruby.yml
@@ -28,9 +28,23 @@ sources:
           \d\.\d\.\d$
 
 conditions:
+  checkForAsdf:
+    kind: shell
+    disablesourceinput: true
+    dependson:
+      - "source#rubyDockerImageLatestVersion"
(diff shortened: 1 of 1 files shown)


dduportal:
Merging as the updatecli diff in CI checks looks good

PR #1526, opened 2024-11-15
chore(jenkinsfile) adds garbage collector as distinct job for azure vms

jayfranco999:
As per https://github.com/jenkins-infra/helpdesk/issues/4355#issue-2586619289

We add the gabrabe collector jenkinsfile as a distinct job in packer-images. The cleanup scripts for aws garbage collector has also been included for the next steps of integrating GC for aws images.

PR #1531, opened 2024-11-18
feat(Jenkinsfile_gc) adds cron trigger expression

jayfranco999:
Related to https://github.com/jenkins-infra/helpdesk/issues/4355#issue-2586619289

As per https://github.com/jenkins-infra/packer-images/pull/1526#pullrequestreview-2438017043

We add cron trigger expression to the pipeline to execute only once daily.

### DIFF
--- a/Jenkinsfile_gc
+++ b/Jenkinsfile_gc
@@ -3,7 +3,7 @@ pipeline {
         label "linux-arm64-docker"
     }
     triggers {
-        cron('H * * * *') // Runs the pipeline once every hour.
+        cron('@daily') // Runs the pipeline once every hour.
     }
     stages {
         stage('Garbage Collection') {
--- a/Jenkinsfile_gc
+++ b/Jenkinsfile_gc
@@ -2,11 +2,11 @@ pipeline {
     agent {
         label "linux-arm64-docker"
     }
+    triggers {
+        cron('H * * * *') // Runs the pipeline once every hour.
+    }
     stages {
         stage('Garbage Collection') {
-            triggers {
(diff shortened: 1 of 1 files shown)


dduportal (changes requested):
- This PR has not been tested: it fails with the following error:

```
08:56:07             WorkflowScript: 6: Unknown stage section "triggers". Starting with version 0.5, steps in a stage must be in a ‘steps’ block. @ line 6, column 9.
08:56:07             stage('Garbage Collection') {
```

=> when changing a pipeline in a PR, start in Draft, validate your change and then turn the PR ready to review once you're sure it is at least parseable (or test the pipeline locally on a custom Jenkins container)

- We want to have the job running once an hour (we never discussed the frequency no worries) instead of once a day ideally. Could you update the PR according to this request?
=> Note as per https://www.jenkins.io/doc/book/pipeline/syntax/#cron-syntax

> In addition, @yearly, @annually, @monthly, @weekly, @daily, @midnight, and @hourly are supported as convenient aliases.

Jenkinsfile_gc
```diff
@@ -2,6 +2,9 @@ pipeline {
     agent {
         label "linux-arm64-docker"
     }
+    triggers {
+        cron('H * * * *') // Runs the pipeline once every hour.
```
dduportal, line 6:
```suggestion
        cron('@daily') // Runs the pipeline once every hour.
```

As per my last comment

jayfranco999: My bad, I missed it. Lets go with the proposed cron trigger expression.

dduportal (approved):

PR #1565, opened 2024-11-28
chore(updatecli): update the updatecli manifest for the correct goss key stdout and not stderr anymore

smerle33:
as updatecli output it's version to stout now, we need to update our manifest.
Ref https://github.com/updatecli/updatecli/pull/2997


jayfranco999:
LGTM! 🚀

PR #1603, opened 2024-12-15
fix(ubuntu/windows) avoid tragedy of commons goods: use `dlcdn` apache download server for Maven

dduportal:
Same as https://github.com/jenkins-infra/github-reusable-workflows/pull/41 + https://github.com/jenkins-infra/github-reusable-workflows/issues/39

We want to avoid the [Tragedy of Common Goods](https://en.wikipedia.org/wiki/Tragedy_of_the_commons) and as such we should only download binaries from the recommended sources.

In the case of Maven, using the Apache CDN instead of archives is the least we can do to decrease our hammering of the Apache public INFRA, and also to decrease chances of failing builds for us.

Note: https://github.com/jenkins-infra/github-reusable-workflows/issues/39#issuecomment-2439796240 shows that there might be "different/better" sources. Worth discussing and changing again, but this PR is a first step.


smerle33 (approved):
💯

PR #1609, opened 2024-12-17
chore(goss/windows): remove retry for goss

smerle33:
to avoid a waste of time and process, we reduce the retry time on windows to the default `0` as per https://github.com/goss-org/goss/blob/00e9355293bbe8b554f8b874cced6a55aa92ffd4/docs/cli.md?plain=1#L341


dduportal (approved):
This change is LGTM

Note: the current tests on Windows shows a timeout on the `hadolint` CLI call which reaches the 10s timeout. This should not happen: might be a subject of another PR if not a temp issue.

PR #1611, opened 2024-12-18
feat(windows) add AWS EC2 builds (2019 and 2022)

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4354

This PRs (re)introduces the Windows (2019 and 2022) builds for AWS EC2.

- Set up Windows AMI in the YAML descriptor in which we track them
  - Added `updatecli` manifest to track them. Will most probably move to another PR, but here for now (failing the Updatecli check)
- Partial revert of https://github.com/jenkins-infra/packer-images/pull/734 (only what we need: source, build block and the sysprep "last" provisionner)
  - Add back the "AWS AMI Windows SysPrep": https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2launch-v2.html
- Add back the Windows AWS EC2 build (and simplify Jenkinsfile matrix)
- Wip: factorize the `amazon-ebs` 's `source_ami` attribute to the source (instead of duplicating per builder). It's a test (now that the `try()` protects us, but might need to be reverted.
- feat: Set up (all!) AWS sources to use enforced IMDS v2 (instead of legacy) to avoid lateral attacks
- cleanup: remove unused `install-packer.sh` script
- Nit: Renamed `local.windows_disk_size_gb` to `local.disk_size_gb`  - downstream PR: https://github.com/jenkins-infra/jenkins-infra/pull/3788
- fix(test): fine tune `goss` on windows to ensure success
  - only check for `hadolint` CLI on Linux to avoid frequent timeouts due to concurency issues
  - Decrease goss concurency


dduportal:
Build 4 did fail with:

- Windows AWS (amazon-ebs) builds are failing with a `Timeout waiting for WinRM` (after retrieving the password successfully from the console)
  - Port is open in the inbound security group, but I guess it is not on the infra.ci agent outbound section. 
- Azure builds are all failing (Ubuntu and Windows) with:

```
 ==> azure-arm.ubuntu: ERROR:   -> InvalidParameter : Resource '<REDACTED>' has invalid parameters. Details: The value of parameter imageReference.version is invalid.
unexpected status 400 (400 Bad Request) with error: InvalidTemplateDeployment: The template deployment '<REDACTED>' is not valid according to the validation procedure.
```

- Docker (Linux) are successful though

PR #1645, opened 2025-01-13
fix(azcopy): install from repository

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4471
AzureEdge.net CDN shuts down was before 15 of january, let's switch to an apt-get install by the repository of microsoft.


smerle33:
I forgot to remove the goss test for azcopy version, I will push the change when the PR to bump git will be merged !

PR #1647, opened 2025-01-13
chore(azcopy/linux): add a version to the apt install of azcopy with the corresponding updatecli manifest

smerle33:
following #1645 and as per https://github.com/jenkins-infra/helpdesk/issues/4471
this PR lock azcopy install to as specific version and track it with updatecli


dduportal (approved):
`Checks are passing: LGTM`

PR #1648, opened 2025-01-13
chore(goss/build-jenkins-agent-windows.pkr) move `max-concurrent` parameter in front

smerle33:
to be sure it's taken in account

as per https://github.com/goss-org/goss/blob/b563479df886927e0bbcbe62dea245ac12d9fb99/cmd/goss/goss.go#L33


smerle33:
not correct

PR #1649, opened 2025-01-13
fix(windows) ensure Git has autoCRLF enabled

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3865

This PR applies @jtnord's proposal in https://github.com/jenkins-infra/pipeline-library/pull/899#discussion_r1910892747 to set up tje default autoCRLF for Git.

Note: it was explicitly set to `false` since 5 years (https://github.com/jenkins-infra/packer-images/pull/8). Not sure why, so let's go with the new setting to fix spotless issues *also* in the Windows VM templates


dduportal:
Merging as the only failure is due to a flaky goss timeout issue

PR #1700, opened 2025-01-29
chore(publish): skip image/ami creation if not on production

smerle33:
to avoid creating AMI or Azure IMAGE if not on production, will save time on PR checks

### DIFF
--- a/sources.pkr.hcl
+++ b/sources.pkr.hcl
@@ -1,6 +1,6 @@
 # This source defines all the common settings for any AWS AMI (whatever Operating System)
 source "amazon-ebs" "base" {
-  skip_create_ami = var.build_type == "prod" ? false : true # skip ami creation if not on prod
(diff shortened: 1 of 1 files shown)


dduportal (commented):
There are 3 types of builds (check https://github.com/jenkins-infra/packer-images/blob/1f70e7d596f4e913f833ad53353df0ce23bae909/Jenkinsfile_k8s#L143 to see how the Terraform variable is defined):

- `prod` is when the build was triggered by a git tag
- `staging` is when the build is triggered on the main branch 
- `dev` for the others (PR, local tests, etc.).


Your change works as expected (the PR did skip successfully the AMI and Azure image creation, decrerasing the effective build time to 51 min instead of > 1h10).
BUT we want to have AMIs created on "staging" (e.g. on build of the main branch) to allow testing new AMIs

dduportal (approved):

PR #1779, opened 2025-03-05
feat(windows): enable LongPaths in registry

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4574

setting the right value within the registry

### DIFF
--- a/build-jenkins-agent-windows.pkr.hcl
+++ b/build-jenkins-agent-windows.pkr.hcl
@@ -94,9 +94,7 @@ build {
   provisioner "powershell" {
     elevated_user     = local.windows_winrm_user[var.image_type]
     elevated_password = build.Password
-    inline = [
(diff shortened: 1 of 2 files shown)


build-jenkins-agent-windows.pkr.hcl
```diff
@@ -90,6 +90,13 @@ build {
     pause_before = "1m"
   }
 
+  ## enable longpath
+  provisioner "powershell" {
+    elevated_user     = local.windows_winrm_user[var.image_type]
+    elevated_password = build.Password
+    inline = ["Set-ItemProperty -Path 'HKLM:\\SYSTEM\\CurrentControlSet\\Control\\FileSystem' -Name LongPathsEnabled -type DWord -Value 1"]
```
dduportal, line 97: This line breaks the packer syntax (see the errors in the `packer init` stage) because the `\\` is an escape character for a litteral slash in terraform (ref. https://developer.hashicorp.com/terraform/language/expressions/strings)

dduportal (approved):

PR #1925, opened 2025-05-13
chore(updatecli) fix golangci-lint manifest

dduportal:
fixup of #1923

Ref. https://github.com/jenkins-infra/packer-images/pull/1924#issuecomment-2875175921


dduportal:
`* key "$.command.golangci-lint.stdout[0]" should be updated from "1.55.2" to "2.1.6", in file "./tests/goss-linux.yaml"`

Let's roll

PR #1932, opened 2025-05-15
fix(ubuntu) use `azcopy` debian package from GH release to circumvent Microsoft Azcopy package errors

dduportal:
Since 48h, the installation of `azcopy` fails with the following error:

```bash
$ apt-get install --yes --no-install-recommends azcopy=10.29.0
Reading package lists...
azure-arm.ubuntu: Building dependency tree...
Reading state information...
Package azcopy is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or is only available from another source
E: Version '10.29.0' for 'azcopy' was not found
```

The package 10.29.0 is not available anymore in the repository. It worked the day before but not anymore.

This PR is a fix which uses the debian packages from the AzCopy GitHub releases (which we use as source for the updatecli version tracking) which are available for both x86_64 and arm64.

We loose the checksum check of the download (but it's HTTPS so not a security issue, more a safety issue of corrupted file) but it's a tradeoff.

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -458,13 +458,15 @@ function install_azurecli() {
 
 ## Ensure that azcopy is installed
 function install_azcopy() {
-  rep_config_pkg="$(mktemp)"
-  # Download and install the repository configuration package.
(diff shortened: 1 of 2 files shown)


smerle33 (approved):
LGTM 🚀

PR #1937, opened 2025-05-15
Add Server-Media-Foundation as part of the Windows provisioning script

lewisbirks:
Adding the Server-Media-Foundation functionality to remove warnings from the playwright testing framework

See https://github.com/jenkinsci/pipeline-graph-view-plugin/pull/761#issuecomment-2883249548

### DIFF
--- a/provisioning/windows-provision.ps1
+++ b/provisioning/windows-provision.ps1
@@ -65,6 +65,9 @@ Function AddToPathEnv($path) {
     Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $newPath | Out-Null
 }
 
+# Requried by testing libraries such as playwright
+Install-WindowsFeature Server-Media-Foundation
+
 # Install OpenSSH (from Windows Features)
 Write-Output "= Installing OpenSSH Server..."
 Add-WindowsCapability -Online -Name OpenSSH.Server~~~~0.0.1.0


provisioning/windows-provision.ps1
```diff
@@ -65,6 +65,9 @@ Function AddToPathEnv($path) {
     Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $newPath | Out-Null
 }
 
+# Requried by testing libraries such as playwright
+Install-WindowsFeature Server-Media-Foundation
```
lewisbirks, line 69: I'm not 100% convinced this is the right place for it, the [documentation](https://support.microsoft.com/en-us/topic/media-feature-pack-list-for-windows-n-editions-c1c6fffa-d052-8338-7a79-a4bb980a700a) on the feature states that some hardware drivers might need re-installing which I'm not sure would happen after this script is run.
Happy to be told otherwise or guided on how best to do it, or to just not do it in general, as the tests appear fine other than the warning appearing at the moment but it might limit the types of tests being written depending on what functionality is required by the Playwright testing framework.

lewisbirks:
@timja how would we go about testing to make sure that this resolves the warning?

timja:
> @timja how would we go about testing to make sure that this resolves the warning?

deploying it is probably the easiest

timja (approved):
Any concerns @dduportal ?

dduportal (approved):
Thanks for the PR!

The checks are green: builds on both Windows 2019 and Windows 2022 showed the same outcome in the logs:

```
Success Restart Needed Exit Code      Feature Result
------- -------------- ---------      --------------
True    No             Success        {Media Foundation}
```

Let's assume it works and perform + deploy a release and we'll see the results

PR #1952, opened 2025-05-20
feat(ubuntu/provision): use temurin api for url and checksum for jdks installations

smerle33:
WIP

as per https://github.com/jenkins-infra/helpdesk/issues/4640
use the api from temurin to set the url and checksum for installation,

PR to follow with updatecli manifest update.

### DIFF
--- a/locals.pkr.hcl
+++ b/locals.pkr.hcl
@@ -31,6 +31,18 @@ locals {
   }
   # Must be greater than 127 Gb to allow Azure template for Windows
   disk_size_gb = 150
+
+  jdk_infos = yamldecode(file("jdks-infos.yaml"))
+  # Mapping architecture to temurin jdk defaults, see https://api.adoptium.net/q/swagger-ui/#/Types/getArchitectures
+  mapped_arch = lookup({
+    "amd64"  = "x64"
+    "arm64"  = "aarch64"
+  }, var.architecture, "x64") # fallback x64
+  # lets list the majors version of jdk to install per architectures
+  jdks = sort([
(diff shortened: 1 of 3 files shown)


locals.pkr.hcl
```diff
@@ -31,6 +31,18 @@ locals {
   }
   # Must be greater than 127 Gb to allow Azure template for Windows
   disk_size_gb = 150
+
+  jdk_infos = yamldecode(file("jdks-infos.yaml"))
+  # Mapping architecture to temurin jdk defaults, see https://api.adoptium.net/q/swagger-ui/#/Types/getArchitectures
+  mapped_arch = lookup({
+    "amd64"  = "x64"
+    "arm64"  = "aarch64"
+  }, var.architecture, "x64") # fallback x64
```
dduportal, line 40: I would keep using the `amd64` and `arm64` values in the YAML. The fact thet Temurin utilizes other platform strings is not a concern to us since we only provide URLs.

Adding this mapping here adds complexity for no benefit (unless I missed something).

provisioning/ubuntu-provision.sh
```diff
@@ -310,57 +310,30 @@ function install_git_gitlfs() {
...
-  rm -f "${temp_archive}"
-}
+## usage of jdks_infos.yaml
+function install_jdks(){
+  for major_jdk_version in ${JDKS}; do
+    echo "jdk to install : ${major_jdk_version}"
+
+    # download jdk and extract
+    installer_url_var="jdk${major_jdk_version}_installer_url"
+    temp_archive="/tmp/jdk${major_jdk_version}.tgz"
```
dduportal, line 320:
```suggestion
    temp_archive="$(mktemp)"
```

nit

PR #1998, opened 2025-06-11
feat: add JDK25 version `25+26-EA`

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4641

introducing jdk25 in agents templates

### DIFF
--- a/updatecli/values.yaml
+++ b/updatecli/values.yaml
@@ -7,7 +7,9 @@ github:
   repository: "packer-images"
 
 jdks:
-  - major: 8
-  - major: 11
-  - major: 17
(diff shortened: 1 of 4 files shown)


updatecli/values.yaml
```diff
@@ -11,3 +11,5 @@ jdks:
   - major: 11
   - major: 17
   - major: 21
+  - major: 25
+    releasetype: ea
```
dduportal, line 15: WDYT about tracking with updatecli on a second PR to make sure we do not miss anything with the first one?

dduportal (approved):

PR #2195, opened 2025-09-02
feat(pipeline) adding windows 2025 to the build list

smerle33:


lemeurherve (approved):
> we will probably wait before merging for windows 2019 to be deprecated in the pipeline

Why? (Asking as I'm eager to test Windows 2025 for building official Windows docker images from a single host)

PR #2489, opened 2026-02-04
windows: Update to Visual Studio Build Tools stable version

slide:
For versions of Windows > 2019, update Visual Studio Build Tools to the stable version.


lemeurherve:
> It looks like the CI passed. Do we have a way of testing before it is deployed?

No, but then the installation on Windows 2019 hasn't changed, should still be good, and there won't be any other consumers except you if you open a PR to change the agent template in jenkinsci/winp when this new packer image will have been released and deployed.

I'd say we're good to go.

provisioning/visualstudio.legacy.vsconfig
```diff
@@ -0,0 +1,23 @@
...
+    "Microsoft.VisualStudio.Component.VC.Redist.14.Latest",
+    "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
+    "Microsoft.VisualStudio.Component.VC.v141.x86.x64",
+    "Microsoft.VisualStudio.Component.VC.Tools.ARM64",
+    "Microsoft.VisualStudio.Component.VC.v141.ARM64",
+    "Microsoft.VisualStudio.Component.VC.Tools.ARM",
+    "Microsoft.VisualStudio.Component.VC.v141.ARM",
+  ],
+  "extensions": []
+}
```
lemeurherve:
```suggestion
}

```
Missing EOL

PR #2638, opened 2026-04-15
feat(Jenkinsfile_k8s): publish build status report on release tag builds

jayfranco999:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2843

Adds a `publishBuildStatusReport()` stage after the Docker manifest build on release tag builds, enabling staleness monitoring for this job.

Gated by `env.TAG_NAME` so only release builds publish status reports.


dduportal (approved):
👏

PR #2687, opened 2026-05-01
fix(windows) ensure Docker Buildx is installed in a valid location

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/5088

Since #2349 which bumped Docker CE from 28.x to 29.x, the system-level plugins installation directiory we were using is deprecated and not checked anymore, leading to the error `docker: unknown command: docker buildx`.

Ref. https://docs.docker.com/engine/release-notes/29/#deprecations-1

> Remove `%PROGRAMDATA%\Docker\cli-plugins` from the list of paths used for CLI plugins on Windows. This path was present for backward compatibility with old installation, but replaced by `%ProgramFiles%\Docker\cli-plugins`. [docker/cli#6713](https://github.com/docker/cli/pull/6713)

This PR changes the installation directory to `${env:ProgramFiles}\Docker\cli-plugins` instead.

----

Tested manually on a pipeline replay on ci.jenkins.io with the agent in version 2.100.1:

```
PS C:\Program Files\Docker\cli-plugins> mkdir "$env:ProgramFiles\Docker\cli-plugins"
PS C:\Program Files\Docker\cli-plugins> cd "$env:ProgramFiles\Docker\cli-plugins"
PS C:\Program Files\Docker\cli-plugins> $ProgressPreference = 'SilentlyContinue' # Disable Progress bar for faster downloads
PS C:\Program Files\Docker\cli-plugins> Invoke-WebRequest "https://github.com/docker/buildx/releases/download/v0.33.0/buildx-v0.33.0.windows-amd64.exe" -Outfile docker-buildx.exe
PS C:\Program Files\Docker\cli-plugins> dir
    Directory: C:\Program Files\Docker\cli-plugins

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----          5/1/2026   8:24 AM       65599488 docker-buildx.exe

PS C:\Program Files\Docker\cli-plugins> docker buildx info
Usage:  docker buildx [OPTIONS] COMMAND

Extended build capabilities with BuildKit

Options:
      --builder string   Override the configured builder instance
  -D, --debug            Enable debug logging

Management Commands:
  dap         Start debug adapter protocol compatible debugger
  history     Commands to work on build records
  imagetools  Commands to work on images in registry
  policy      Commands for working with build policies

Commands:
  bake        Build from a file
  build       Start a build
  create      Create a new builder instance
  dial-stdio  Proxy current stdio streams to builder instance
  du          Disk usage
[...]

PR #2698, opened 2026-05-04
fix(windows): add docker plugin folder to `PATH`

lemeurherve:
Even with #2687, `docker buildx` is still "not found". This change adds docker plugins folder to $PATH to fix it.

Follow-up of:
- #2687
- #2697

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5088


lemeurherve:
Closing as it's not the issue: docker plugin folder isn't properly defined:
> image

PR #2763, opened 2026-05-29
chore(Jenkinsfile_gc) discard builds (max 200 in the history) and avoid concurent executions

dduportal:
Build discarder: ref. https://github.com/jenkins-infra/helpdesk/issues/5145#issuecomment-4575645259


dduportal:
The GC build (not reporting any checks: it's expected) for this PR reports a clean setup:

Capture d’écran 2026-05-29 à 15 35 00


Let's roll

PR #2789, opened 2026-06-09
fix(ubuntu) correct playwright setup and set up headless chromium

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5174

This PR fixes the Playwright setup which was broken by #2775

It simplifies the setup installation by only running the "full" `install-deps` command as `root` user, instead of the former fragile stdout parsing with the `--dry-run` flag (since ASDF as `jenkins` user is not required anymore).

It also pre-installs the `chromium` headless cache as `jenkins` user (tested on a local container) to avoid requiring pipelines to run `npx playwright install` command on each run.

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -599,25 +599,19 @@ function install_nodejs() {
 }
 
 function install_playwright() {
-  # Install pinned playwright globally first (as NPX needs it)
-  source /etc/environment && npm install -g playwright@"${PLAYWRIGHT_VERSION}"
+  source /etc/environment
 
-  # Install required system dependencies - https://playwright.dev/docs/browsers#install-system-dependencies
-  # Note: need to extract the list of missing packages as jenkins user but install as root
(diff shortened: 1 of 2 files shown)


dduportal:
Putting as draft: it only works on `arm64` Ubuntu but not x86_64. Reproducing the issue locally

dduportal:
> Putting as draft: it only works on `arm64` Ubuntu but not x86_64. Reproducing the issue locally

🤦 On `arm64`, the Chrome headless binary is:

```
/home/jenkins/.cache/ms-playwright/chromium_headless_shell-1223/chrome-linux/headless_shell
```

while on `x86_64` it is:

```
/home/jenkins/.cache/ms-playwright/chromium_headless_shell-1223/chrome-headless-shell-linux64/chrome-headless-shell
```

so that's why the test is failing.


Also: NPX should not be used in the provisioning to avoid version mismatches

dduportal:
All green on Linux, finally!

PR #2836, opened 2026-06-18
chore(Ubuntu): change `jq` installation method to binary

jayfranco999:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5117
- https://github.com/jenkins-infra/helpdesk/issues/5184 (ticks the "Move jq as binary installation" task)

jq has been pinned at 1.6 because the version bump and the Linux install mechanism were coupled: `ubuntu-provision.sh` installed jq via `apt-get install jq=${JQ_VERSION}*`, and Ubuntu's repos don't serve 1.8.x. That's why the auto-generated bump PR (#1064) could never merge.

## Notes

- jq publishes a raw binary (not a tarball), so the install uses `curl --output` + `chmod a+x` rather than extracting an archive.
- The goss manifest prefix fix (so future auto-bumps write `jq-<version>` correctly) is in #2835.

## Follow-up (not in this PR)

Two repos still download their own jq and should switch to the agent-provided one once this lands (AC #2 of #[5117](https://github.com/jenkins-infra/helpdesk/issues/5117)): `update-center2/site/publish.sh` (jq 1.5) and `javadoc/scripts/generate-javadoc.sh` (jq 1.6), both from the deprecated`stedolan/jq`. Also revert the 1.6 syntax workaround from #2733.

### DIFF
--- a/provisioning/tools-versions.yml
+++ b/provisioning/tools-versions.yml
@@ -23,7 +23,7 @@ helm_secrets_version: 4.7.7
 helm_version: 3.19.1
 helmfile_version: 1.5.5
 jenkins_remoting_version: 3355.v388858a_47b_33
-jq_version: 1.6
+jq_version: 1.8.1
 jxreleaseversion_version: 2.9.3
 kubectl_version: 1.34.9
 launchable_version: 1.124.4
--- a/provisioning/tools-versions.yml
+++ b/provisioning/tools-versions.yml
@@ -30,7 +30,7 @@ launchable_version: 1.124.4
 maven_version: 3.9.16
 netlifydeploy_version: 0.1.8
 nodejs_linux_version: 24.18.0
-nodejs_windows_version: 24.17.0
+nodejs_windows_version: 24.18.0
 nuget_version: 7.6.0
 openssh_authorized_keys_url: https://raw.githubusercontent.com/jenkins-infra/aws/main/ec2_agents_authorized_keys
 packer_version: 1.15.4
--- a/provisioning/tools-versions.yml
+++ b/provisioning/tools-versions.yml
@@ -1,10 +1,10 @@
 # if manually change, please update the goss file accordingly
 asdf_version: 0.15.0
-awscli_version: 2.35.8
+awscli_version: 2.35.11
 azcopy_version: 10.32.4
 azurecli_version: 2.87.0
 chocolatey_version: 1.4.0
-compose_version: 5.1.4
+compose_version: 5.2.0
 cst_version: 1.22.1
 docker_buildx_version: 0.35.0
 docker_version: 29.6.0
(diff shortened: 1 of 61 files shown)


provisioning/tools-versions.yml
```diff
@@ -23,7 +23,7 @@ helm_secrets_version: 4.7.7
 helm_version: 3.19.1
 helmfile_version: 1.5.5
 jenkins_remoting_version: 3355.v388858a_47b_33
-jq_version: 1.6
+jq_version: 1.8.1
```
lemeurherve:
```suggestion
jq_version: 1.6
```
I would do that upgrade in a follow-up PR, or merge #1064 instead.

This pull request should be dedicated to install `jq` from GitHub release assets only. (separation of concerns)

provisioning/ubuntu-provision.sh
```diff
@@ -229,16 +229,22 @@ function install_golangcilint(){
     tar --extract --gunzip --strip-components=1 --directory="/usr/local/bin/" "golangci-lint-${GOLANGCILINT_VERSION}-linux-${ARCHITECTURE}/golangci-lint"
 }
 
+## Install jq from the official GitHub release (single source, OS-independent)
+function install_jq(){
+  # JQ_VERSION is an env var provided outside of the script
+  # shellcheck disable=SC2153
```
lemeurherve, line 235:
```suggestion
```
First line unneeded, second one should **eventually** be applied to the whole file as it concerns almost all `install_*` functions. (better yet, set a default empty value to satisfy shellcheck, ex: `${JQ_VERSION:-}`)

dduportal: This suggestion should be treated in a separate PR

provisioning/tools-versions.yml
dduportal: +1 with @lemeurherve : let's rename/update this PR to focus only on the installation method change. And then  we'll let updatecli do it's magic for the up
date

lemeurherve: @jayfranco999 let's roll with the bump included in this PR.

Small nit before merging, can you remove `(jq)` from the pull request title please?

dduportal: +1 with the all in one PR (I dion't have a strong opininion on both :D)

dduportal:
Let's keep this PR "all in one". However, the title should be updated:

- Removing the `(jq)` scope (usual scopes are: `windows`, `ubuntu`, `Jenkinsfile`, `GHA` or `updatecli`) as pointed by @lemeurherve in https://github.com/jenkins-infra/packer-images/pull/2836#discussion_r3441330272
- To keep it consistent with https://github.com/jenkins-infra/packer-images/blob/b46611d97e17c7e834671cebc2464e71eb240422/updatecli/updatecli.d/jq.yml#L63, WDYT about replacing `install jq 1.8.1` by `Bump JQ version to 1.8.1`?
- Since Windows installation method is not change, I suggest updating the end of the title to specify it's ubuntu only (`and change Ubuntu installation method to binary`)

lemeurherve:
@jayfranco999: `jq` is installed on Windows too, suggesting to remove `(ubuntu)` from the scope of your PR title ;)

dduportal:
Ping @jayfranco999 

- Can you rebase this PR (and fix conflicts)
- Take in account last feedback from Herve regarding title?

lemeurherve (approved):

tests/goss-common.yaml
```diff
@@ -63,7 +63,7 @@ command:
     exec: jq --version
     exit-status: 0
     stdout:
-      - jq-1.6
+      - jq-1.8.1
```
lemeurherve:
```suggestion
      - jq-1.6
```
As this PR doesn't include any bump anymore, as its current version is 1.6 cf https://github.com/jenkins-infra/packer-images/blob/d095b24ddbfb7f5a50671d51af5ca51549867d2c/provisioning/tools-versions.yml#L26

and as https://github.com/jenkins-infra/packer-images/pull/1064 bumps it to 1.8.2 since one hour, I'd suggest editing out "bump jq to 1.8.1 and " from this PR title, WDYT?

dduportal (approved):

tests/goss-common.yaml
dduportal: ~I don't understand your suggestion: the PR here **does** bump JQ from 1.6 to 1.8.1~

dduportal: Oh no, nevermind, I was watching the wrong version of the code. LGTM with the suggestion: this PR is back to "only" changing the installation method on Linux. I lost track

lemeurherve (approved):

dduportal (approved):
- New title looks good to me
- Root cause of today's changes comes from the conflict resolution which accidentally rolled back the `1.8.1` specified version we all agreed on to `1.6`
- Checked with @jayfranco999 , title re-updated to map to the current change (+ fixed the feedback around test by @lemeurherve ) 


Agreement: this PR does NOT change `jq` version anymore. Will be done to 1.8.2

lemeurherve:
Last post merge title edit to include what has been changed (`jq`).

Issue #2824, opened 2026-06-17
Consider conditional builds on PRs


dduportal:
Following up https://github.com/jenkins-infra/helpdesk/issues/5177, @lemeurherve suggested that we could have conditional builds for PRs to limit the amount of machines.

Costs are not that high (at least not due to PRs building everything) but it could come handy to tighten our feedback loop when working on Linux features/bug fixes for instance.

Proposal for the first step: per-OS build

- Each PR build (and only them) has a pipeline step which gathers labels (using `gh` CLI?)
- For each OS from the list (`ubuntu-24.04`, `windows-2019`, `windows-2022`, `windows-2025`), if there is a label with the corresponding  string then we add them to the list
- If there is a `windows` label, we build all windows (and ignore all windows-XXX labels)
- If there is an `ubuntu` label, then we build all Ubuntu (and ignore all ubuntu-XXX labels) 

WDYT @lemeurherve ?

lemeurherve:
Implemented in https://github.com/jenkins-infra/packer-images/pull/2839

dduportal:
Initial implementation looks god regarding the OS classes.

However it still have a major problem to solve: we are missing the "per cloud provider" condition. It's serious enough to justify not closing the issue here, but not serious enough to "rush" something.

See https://github.com/jenkins-infra/packer-images/pull/2911 which only bumps the base image for Windows 2019 on `AWS` only.
=> Its build did only build the *Azure* image, not the AWS one:

<img width="1290" height="433" alt="Image" src="https://github.com/user-attachments/assets/45af49a2-da8f-4a47-ace3-1eab58d74500" />
<img width="608" height="278" alt="Image" src="https://github.com/user-attachments/assets/23c44b83-932b-4732-a6ee-7cd30c559729" />

Somehow it is expected because we have not implemented any conditional logic on the provider so it defaults to the "partial matrix" we've implemented where Windows 2019 is only tested on Azure on PRs: it's a WIndows 2019 only PR so the behavior makes sense.

We should implement a case where, if the provider label (might need to revisit provider labels from updatecli so they map to packer's providers but it's easy) is specified then "force" the aforementioned provider in the build matrix

dduportal:
- Per provider conditional builds implemented in https://github.com/jenkins-infra/packer-images/pull/2922
- Next step: per CPU architecture

PR #2846, opened 2026-06-19
chore(windows): skip Windows updates and AMI preparation on pull requests

lemeurherve:
This change skips the `windows-updates` provisioners and the final `windows-restart` and `sysprep` steps on pull requests, executed only when the `build_type` is not `dev` (on primary branch).

Related:
- #2824

### DIFF
--- a/build-jenkins-agent-windows.pkr.hcl
+++ b/build-jenkins-agent-windows.pkr.hcl
@@ -2,6 +2,11 @@ local "common_goss_args" {
   expression = "--use-alpha=1 --loglevel DEBUG validate --max-concurrent=1 --retry-timeout 60s --sleep 60s --format documentation"
 }
 
+locals {
+  skip_on_pr = var.build_type == "dev"
+  skip_on_pr_except_for_2019 = var.build_type == "dev" && var.agent_os_version != "2019"
+}
+
 build {
   source "amazon-ebs.base" {
     name           = "windows"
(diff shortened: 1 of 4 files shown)


build-jenkins-agent-windows.pkr.hcl
```diff
@@ -2,6 +2,10 @@ local "common_goss_args" {
   expression = "--use-alpha=1 --loglevel DEBUG validate --max-concurrent=1 --retry-timeout 60s --sleep 60s --format documentation"
 }
 
+locals {
+  is_pullrequest = var.change_id != ""
+}
```
dduportal: You could use the existing variable `build_type` which defaults to `dev` when not on the primary branch or not a tag (see https://github.com/jenkins-infra/packer-images/blob/6d0fe338467559acd17f6b7ddaa141cb0e5ebfbe/Jenkinsfile_k8s#L169).

=> Yes it covers a bit more than just pull requests, but the really rare cases where we create a branch build with no PRs, I expect the value set to `staging` manually by the branch author, otherwise it  is the same as a PR

build-jenkins-agent-windows.pkr.hcl
```diff
@@ -2,6 +2,10 @@ local "common_goss_args" {
   expression = "--use-alpha=1 --loglevel DEBUG validate --max-concurrent=1 --retry-timeout 60s --sleep 60s --format documentation"
 }
 
+locals {
+  is_pullrequest = var.change_id != ""
```
lemeurherve:
```suggestion
  is_pullrequest = var.build_type == "dev"
```

dduportal (commented):
Is this PR ready to roll?

lemeurherve:
> Is this PR ready to roll?

The 3 "windows-update" are executed even with the `only` conditions, I still need to find out why.

I'll put this PR in ready for review when ready to roll 🙂 

EDIT: got it working by putting a dummy content instead of `[]`.

It's cutting almost a third of build time on Azure Windows 2022 🎉

lemeurherve:
Added a specific `skip_on_pr_except_for_2019` local in https://github.com/jenkins-infra/packer-images/pull/2846/commits/6ededeaad12dc0fd381aa9bd1a67c9171ec9c4ed as on the first two builds OpenSSH setup failed. It seems to require those Windows updates.

build-jenkins-agent-windows.pkr.hcl
```diff
@@ -2,6 +2,11 @@ local "common_goss_args" {
   expression = "--use-alpha=1 --loglevel DEBUG validate --max-concurrent=1 --retry-timeout 60s --sleep 60s --format documentation"
 }
 
+locals {
```
lemeurherve: Moving those locals to locas.pkr.hcl, just waiting for the current build to show if `skip_on_pr_except_for_2019` works as intended before pushing this change.

build-jenkins-agent-windows.pkr.hcl
lemeurherve, line 67: We could "gain" a restart by installing Docker CE before the `windows-update` provisioners.

WDYT?

dduportal (approved):

build-jenkins-agent-windows.pkr.hcl
dduportal: You might want to re-read the comment on the line right before :D

dduportal: (The comment may be updated though)

build-jenkins-agent-windows.pkr.hcl
lemeurherve: Amazon hasn't finished yet, usually taking about 1 hour to complete, pushing my change.

The others are not bad 🙂 

> image

PR #2847, opened 2026-06-19
Revert "feat(ubuntu): install `python3` from source and bump to 3.14.6"

lemeurherve:
Reverts jenkins-infra/packer-images#2838

See https://github.com/jenkins-infra/packer-images/pull/2838#issuecomment-4753042211 (missing module(s))

PR #2844, opened 2026-06-19
chore(updatecli): bump `python3` version in goss-linux.yml too

lemeurherve:
Follow-up of:
- #2838 

Ref:
- https://github.com/jenkins-infra/packer-images/issues/2837

### Testing done

<details><summary>updatecli diff --config updatecli/updatecli.d/python3.yml --values updatecli/values.yaml</summary>

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/python3.yml"

SCM repository retrieved: 1


++++++++++++++++++
+ AUTO DISCOVERY +
++++++++++++++++++



++++++++++++
+ PIPELINE +
++++++++++++


########################
# BUMP PYTHON3 VERSION #
########################

Pipeline ID     : 2069c037197af68ddd6a68a8065530ff3a4c40696f8f57f39cd43a0d3472bd10
Dry Run         : enabled

source: lastReleaseVersion
--------------------------

The shell 🐚 command "/bin/sh /var/folders/tp/wg7m13056jjgb2z1_jv07j1r0000gn/T/updatecli/bin/52fefa4e0fd6ab22d2a3ff96eb321e386acf13641ea5fa56ee15ac187195e27b.sh" ran successfully with the following output:
----
3.14.6
----
✔ shell command executed successfully

condition: checkForChocolateyPackage
------------------------------------

The shell 🐚 command "/bin/sh /var/folders/tp/wg7m13056jjgb2z1_jv07j1r0000gn/T/updatecli/bin/4c45b0bb49cac15d7025256bfd3ef03723c8409d268504fd9963ddc22a3b8738.sh" ran successfully with the following output:
----
                        <div class="callout-header">Package Approved</div>
----
✔ shell condition of type "console/output", passing

condition: checkPythonSourceAvailable
-------------------------------------

The shell 🐚 command "/bin/sh /var/folders/tp/wg7m13056jjgb2z1_jv07j1r0000gn/T/updatecli/bin/2dfc6987508dd7bfc014d2ad90cec48664da1c5734b6435282b27025f94f365c.sh" ran successfully with the following output:
----
HTTP/2 200 
server: nginx
last-modified: Wed, 10 Jun 2026 10:49:05 GMT
etag: "6a294121-1dc9a44"
x-clacks-overhead: GNU Terry Pratchett
content-type: application/octet-stream
via: 1.1 varnish, 1.1 varnish, 1.1 varnish
accept-ranges: bytes
age: 770980
[...]

PR #2838, opened 2026-06-18
feat(ubuntu): install `python3` from source and bump to 3.14.6

lemeurherve:
This change builds and installs `python3` from source so we can have the version we want (cf tool-versions).

I added the minimal amount of dev dependencies, we might need to add some later.

Ref:
- https://github.com/jenkins-infra/packer-images/issues/2837
- https://github.com/jenkins-infra/helpdesk/issues/4714

Issue #2837, opened 2026-06-18
Install `python3` from source


lemeurherve:
### What feature do you want to see added?

As noted in #2830, the version of python3 available in Ubuntu is dependant of the release line.

In order to use Ansible for testing instead of `goss`, we need at least the version 3.12, which isn't available in Ubuntu 22.04.

To decouple `python3` from Ubuntu release line, we should install it from source, at least on Linux for now[^1].

Refs:
- https://docs.python.org/3/using/unix.html#building-python
- https://devguide.python.org/getting-started/setup-building
- https://github.com/jenkins-infra/helpdesk/issues/4714

[^1]: We can install whichever python version we want with Chocolatey, keeping it simple while we are not getting rid of Chocolatey (yet).

lemeurherve:
Reopening, missing modules: https://github.com/jenkins-infra/packer-images/pull/2838#issuecomment-4753042211

lemeurherve:
Got a successful build of jenkinsci/packaging using python3 version 3.14 and ansible 13.0.0: https://github.com/jenkinsci/packaging/pull/727#issuecomment-4791307856

lemeurherve:
Note: the jenkins-infra/docker-packaging also needs python3 version 3.14 to be installed, currently at 3.12.3 (cf replay https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-727/7/stages/?start-byte=0&selected-node=20#log-20-1)

Switching it from Ubuntu:24.04 to 26.04 (ref: https://github.com/jenkins-infra/helpdesk/issues/5124) would allow to get python3.14 OOTB in the `build` stage of jenkinsci/packaging: https://github.com/jenkinsci/packaging/blob/c81721a781ce6dbb7e2f76eb2f5240ac292d3e8c/Jenkinsfile#L16-L42

lemeurherve:
> Note: the jenkins-infra/docker-packaging also needs python3 version 3.14 to be installed, currently at 3.12.3 (cf replay https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-727/7/stages/?start-byte=0&selected-node=20#log-20-1)
> 
[...]

PR #2848, opened 2026-06-19
fix(ubuntu): don't simlink `python3` built from source to `/usr/local/bin/python3`

lemeurherve:
This change allows to keep both python3 installed from Ubuntu/debian package (3.10) and python3 installed from source (3.14) so I can continue tinkering and progressing on a proper installation from source with all required modules, without reverting it altogether (cf https://github.com/jenkins-infra/packer-images/pull/2847, which would also require reverting https://github.com/jenkins-infra/packer-images/pull/2844).

Refs:
- https://github.com/jenkins-infra/packer-images/pull/2838#issuecomment-4753042211
- https://github.com/jenkins-infra/packer-images/issues/2837

### Testing done

```bash
$ docker run -it --rm jenkinsciinfra/jenkins-agent-ubuntu-22.04 /bin/bash        
jenkins@8e728ee88227:~$ whereis python3
python3: /usr/bin/python3 /usr/lib/python3 /etc/python3 /usr/share/python3 /opt/az/bin/python3
jenkins@8e728ee88227:~$ python3 --version
Python 3.10.12
jenkins@8e728ee88227:~$ /usr/local/bin/python3.14 --version
Python 3.14.6
```

### DIFF
--- a/tests/goss-linux.yaml
+++ b/tests/goss-linux.yaml
@@ -90,10 +90,16 @@ command:
   parallel:
     exec: parallel --version
     exit-status: 0
-  python3:
+  # Python3 installed from Ubuntu/debian package
+  python3package:
     exec: python3 --version
     exit-status: 0
-    # TODO: track with updatecli
+    stdout:
+      - "3.10"
+  # Python3 installed from source
+  python3:
+    exec: /usr/local/bin/python3.14 --version
+    exit-status: 0
     stdout:
       - 3.14.6
   rsync:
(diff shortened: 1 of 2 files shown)


tests/goss-linux.yaml
```diff
@@ -90,10 +90,16 @@ command:
...
     exec: parallel --version
     exit-status: 0
-  python3:
+  # Python3 installed from Ubuntu/debian package
+  python3package:
     exec: python3 --version
     exit-status: 0
-    # TODO: track with updatecli
+    stdout:
+      - "3.10"
```
lemeurherve, line 98: Specifying only a minor version and not a patch as we don't know when this package can be updated in the base image exactly.

Also, in quote to avoid YAML's float convertion gotcha.

dduportal (approved):

Issue #1731, opened 2025-02-12
Re-enable Windows test harness


dduportal:
https://github.com/jenkins-infra/packer-images/pull/1699 provided a quick fix on Windows which need to be reverted

dduportal:
Still flaky: @lemeurherve is working on a PoC to use Ansible instead of Goss

PR #2865, opened 2026-06-25
feat(linux): add `ansible`

lemeurherve:
Ref:
- https://github.com/jenkins-infra/packer-images/issues/1731#issuecomment-4732170211

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -661,6 +661,7 @@ function install_ansible() {
   python3.14 -m venv "${ansible_venv_dir}"
   . "${ansible_venv_dir}"/bin/activate
   pip --require-virtualenv --no-cache-dir install ansible-core=="${ANSIBLE_CORE_VERSION}"
+  deactivate
 }
 
 ## ensure yamllint is installed
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -35,6 +35,7 @@ groupid=1001
 install_dir=/usr/local/bin
 asdf_install_dir="${userhome}/.asdf"
 launchable_venv_dir="/usr/local/launchable"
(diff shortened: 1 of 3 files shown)


tests/goss-linux.yaml
```diff
@@ -98,7 +103,7 @@ command:
       - "3.10"
   # Python3 installed from source
   python3:
-    exec: /usr/local/bin/python3.14 --version
+    exec: python3.14 --version
```
dduportal: Movable to common tests?

lemeurherve: Cf https://github.com/jenkins-infra/packer-images/pull/2844#pullrequestreview-4532559333

> (🤦 the binary is `python` on Windows but `python3` on Linux so we can't use the common tests)

Still the same situation here unfortunately.

---
Note: as this change was unrelated to this PR, I've rebased it out.

dduportal: > Still the same situation here unfortunately.

I will learn, I promise, I will learn 😶‍🌫️

provisioning/ubuntu-provision.sh
```diff
@@ -655,6 +656,13 @@ function install_launchable() {
   ln -s "${launchable_venv_dir}/bin/launchable" /usr/local/bin/launchable
 }
 
+## Ensure ansible is installed
+function install_ansible() {
+  python3.14 -m venv "${ansible_venv_dir}"
+  . "${ansible_venv_dir}"/bin/activate
+  pip --require-virtualenv --no-cache-dir install ansible-core=="${ANSIBLE_CORE_VERSION}"
+}
```
dduportal, line 665: I don't remember the exact command to do it, but I suggest we disable the venv once the installation is complete. Because in shell, even exiting the current function won't exit the venv itself.

So if any python-related installation happens after this, it will be installed in the ansible venv accidentally

dduportal (approved):

PR #2869, opened 2026-06-26
chore(linux): use `ansible` instead of `goss` for tests

lemeurherve:
WIP, for discussions

Ref:
- #1731

### DIFF
--- a/tests/ansible/playbook.yml
+++ b/tests/ansible/playbook.yml
@@ -1,5 +1,46 @@
 ---
-- name: tests
+# - name: tests
+#   hosts: all
+#   become: true
+#   become_user: jenkins
+#   gather_facts: true
+#   vars_files:
+#     - ../../provisioning/tools-versions.yml
+#     - ../../jdks-infos.yaml
+#   # vars:
+#   #   ansible_python_interpreter: /usr/local/bin/python3.14
+#   #   ansible_remote_tmp: /tmp/.ansible
+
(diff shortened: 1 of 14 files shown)


tests/ansible/tasks/tests_common.yml
```diff
@@ -0,0 +1,162 @@
+---
+- name: Check awscli
+  ansible.builtin.command: aws --version
+  register: awscli_version_output
+  changed_when: false
+  failed_when: "awscli_version not in awscli_version_output.stdout"
+  ignore_errors: true
```
lemeurherve, line 7: Added for now to see which tests don't pass without exiting on the first error.

tests/ansible/tasks/tests_linux.yml
lemeurherve, line 1: First draft from Claude Code, convertion of goss-linux.yml, to be revised.

tests/ansible/playbook.yml
```diff
@@ -0,0 +1,15 @@
...
+  gather_facts: false
+  vars_files:
+    - ../../provisioning/tools-versions.yml
+    - ../../jdks-infos.yaml
+
+  tasks:
+    - name: Import common tests
+      ansible.builtin.import_tasks: tasks/tests_common.yml
+
+    - name: Import Linux tests
```
lemeurherve, line 64: For later: tests_windows.yml, with "windows_only" tag

PR #2892, opened 2026-07-13
chore(ubuntu): Bump helm to 4.2.3 and fix plugin install for Helm 4

jayfranco999:
ref: 
- https://github.com/jenkins-infra/helpdesk/issues/5127

Helm 4 changed two things that break the current plugin install:

- Plugin signatures are verified on install by default, so all three plugin installs now need --verify=false (the sources are unsigned).
- All-in-one plugins are no longer supported. helm-secrets is now split into secrets, secrets-getter and secrets-post-renderer, so the `helm plugin install <repo> --version` form only installs a partial, non-working plugin and the `helm secrets` command disappears.

helm-diff and helm-git still work with the repo + --version form, they just need --verify=false. helm-secrets is installed from its three release artifacts instead.

Testing:

Reproduced the current install in a clean ubuntu:22.04 container with helm 4.2.3. All three installs failed on the verification gate. 

After adding `--verify=false` to the plugins installation:

  ```
  NAME                    VERSION TYPE            APIVERSION  PROVENANCE
    diff                    3.15.10 cli/v1          legacy      unknown
    helm-git                1.5.2   getter/v1       legacy      unknown
    secrets                 4.7.7   cli/v1          v1          signed
    secrets-getter          4.7.7   getter/v1       v1          signed
    secrets-post-renderer   4.7.7   postrenderer/v1 v1          signed

    $ helm secrets --version
    4.7.7

```

Separately verified in the kubernetes-management repo that helm 4 renders charts identically to helm 3 (diffed a real cert-manager release) and that helm-diff and helm-git work under Helm 4.

### DIFF
--- a/provisioning/ubuntu-provision.sh
+++ b/provisioning/ubuntu-provision.sh
@@ -382,9 +382,13 @@ function install_helm(){
     "https://get.helm.sh/helm-v${HELM_VERSION}-linux-${ARCHITECTURE}.tar.gz" | \
     tar --extract --gunzip --strip-components 1 --directory="${install_dir}"/ "linux-${ARCHITECTURE}/helm"
 
-  su - "${username}" -c "helm plugin install https://github.com/databus23/helm-diff --version v${HELM_DIFF_VERSION}"
-  su - "${username}" -c "helm plugin install https://github.com/jkroepke/helm-secrets --version v${HELM_SECRETS_VERSION}"
-  su - "${username}" -c "helm plugin install https://github.com/aslafy-z/helm-git.git --version v${HELM_GIT_VERSION}"
+  # Helm 4 verifies plugin signatures on install by default; these sources are unsigned so verification is skipped.
+  su - "${username}" -c "helm plugin install https://github.com/databus23/helm-diff --version v${HELM_DIFF_VERSION} --verify=false"
+  # Helm 4 dropped all-in-one plugins, so helm-secrets is split into three packages that must be installed from their release artifacts.
+  su - "${username}" -c "helm plugin install https://github.com/jkroepke/helm-secrets/releases/download/v${HELM_SECRETS_VERSION}/secrets-${HELM_SECRETS_VERSION}.tgz --verify=false"
+  su - "${username}" -c "helm plugin install https://github.com/jkroepke/helm-secrets/releases/download/v${HELM_SECRETS_VERSION}/secrets-getter-${HELM_SECRETS_VERSION}.tgz --verify=false"
+  su - "${username}" -c "helm plugin install https://github.com/jkroepke/helm-secrets/releases/download/v${HELM_SECRETS_VERSION}/secrets-post-renderer-${HELM_SECRETS_VERSION}.tgz --verify=false"
+  su - "${username}" -c "helm plugin install https://github.com/aslafy-z/helm-git.git --version v${HELM_GIT_VERSION} --verify=false"
 }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: provisioning/ubuntu-provision.sh [hunk 1]
```diff
@@ -382,9 +382,13 @@ function install_helm(){
     "https://get.helm.sh/helm-v${HELM_VERSION}-linux-${ARCHITECTURE}.tar.gz" | \
     tar --extract --gunzip --strip-components 1 --directory="${install_dir}"/ "linux-${ARCHITECTURE}/helm"
 
-  su - "${username}" -c "helm plugin install https://github.com/databus23/helm-diff --version v${HELM_DIFF_VERSION}"
-  su - "${username}" -c "helm plugin install https://github.com/jkroepke/helm-secrets --version v${HELM_SECRETS_VERSION}"
-  su - "${username}" -c "helm plugin install https://github.com/aslafy-z/helm-git.git --version v${HELM_GIT_VERSION}"
+  # Helm 4 verifies plugin signatures on install by default; these sources are unsigned so verification is skipped.
```
lemeurherve:
```suggestion
  # Helm 4 verifies plugin signatures on install by default; these sources are currently unsigned so verification is skipped.
```

provisioning/ubuntu-provision.sh
```diff
@@ -382,9 +382,13 @@ function install_helm(){
...
+  su - "${username}" -c "helm plugin install https://github.com/databus23/helm-diff --version v${HELM_DIFF_VERSION} --verify=false"
+  # Helm 4 dropped all-in-one plugins, so helm-secrets is split into three packages that must be installed from their release artifacts.
```
lemeurherve:
```suggestion
  # Helm 4 dropped all-in-one plugins, so helm-secrets is now split into three packages
```
remove obvious part

lemeurherve (approved):
LGTM!

Note: I've put the `linux` build to get those images built.

Also, https://github.com/jenkins-infra/packer-images/pull/2351 has the `enhancement` label, I would suggest adding it to that PR and set this as title:

> feat(linux): bump helm from v3 to v4.2.3 and disable verification on plugin installations

jayfranco999:
> Also, #2351 has the `enhancement` label, I would suggest adding it to that PR and set this as title:
> 
> > feat(linux): bump helm from v3 to v4.2.3 and disable verification on plugin installations

Since the upgrade isn't something which adds to the features for the end-users I thought chore was justified.

lemeurherve:
> > Also, #2351 has the `enhancement` label, I would suggest adding it to that PR and set this as title:
> > > feat(linux): bump helm from v3 to v4.2.3 and disable verification on plugin installations
> 
> Since the upgrade isn't something which adds to the features for the end-users I thought chore was justified.

@jayfranco999: excluding "chore/feat", how about the rest of the title change?

dduportal:
> > Also, #2351 has the `enhancement` label, I would suggest adding it to that PR and set this as title:
> > > feat(linux): bump helm from v3 to v4.2.3 and disable verification on plugin installations
> 
> Since the upgrade isn't something which adds to the features for the end-users I thought chore was justified.

I strongly disagree with this stance. The "end-users" (e.g. the consumers of the generated artifacts) are... oursleves in kubernetes-management. This change IS an enhancement and as such requires a title starting with `feat`.


Also +1 with Herve's suggestion: the title of this PR is missing a really important information about the plugin verification which is disabled.

PR #2922, opened 2026-07-21
chore(pipeline) implement conditional build for packer providers when specified through PR labels

dduportal:
Ref. #2824 

This PR implements a logic to conditionally build packer providers on pull requests based on labels:

- Dynamically retrieve the packer provider from `packer-*` labels but there is a validation. Pros: avoid "dynamic injection" from labels. Cons: needs to maintain the list in the pipeline.
- No `packer-` label means "Build all providers" (same behavior as before). Note: this is slightly different than with the `OS` labels where "no labels" means "no build".


dduportal:
Tested a few builds and it work as expected:

- OS + `packer-docker` 
- All OSes + `packer-amazon-ebs`

PR #2942, opened 2026-07-29
fix(windows) use NodeJS LTS package with Chocolatey

dduportal:
Our builds are failing since the past 24h: it seems that the 24.18.0 version of NodeJS has been removed for the Chocolatey package `nodejs.install` we're using: https://community.chocolatey.org/packages/nodejs.install#versionhistory. My bet is on the fact that https://nodejs.org/en/blog/release/v24.18.1 is a security release so using 24.18.0 might be risky (if you are running apps with it which we don't here)

This PR changes the package to `nodejs-lts` for Windows:
- It seems to be more stable regarding versions history: https://community.chocolatey.org/packages/nodejs-lts#versionhistory.
- It makes the "use LTS" intent a bit more clear and easier to keep


dduportal (commented):
- `The install of nodejs-lts was successful.` => main build looks good
- Updatecli shows the `24.18.1` available (https://github.com/jenkins-infra/packer-images/pull/2943)

PR #3037, opened 2026-09-03
chore(ansible): migrate install_jq to an Ansible role

jayfranco999:
Moves `install_jq` out of `provisioning/ubuntu-provision.sh` into an Ansible role, as a first pass at the shell to Ansible migration. One `provision.yml` calling roles, so the next function is one more role plus one more line.

Tool versions come from `var.provision_env_file`, the same `tools-versions.yml` the shell provisioning already uses, passed as `--extra-vars @file`. Nothing changes under `updatecli/`. `architecture` is passed separately since it is a Packer variable, not a tracked version.

Two things worth flagging:

**`Jenkinsfile_k8s`** - `packer validate` shells out to `ansible-playbook` as a preflight, so without it on the PATH every matrix cell fails validation, Windows included. `install_ansible()` puts it in `/usr/local/ansible-venv` then deactivates, so it exists but is not findable. Added that dir next to the asdf shims, plus to the `linux-amd64-docker` node in `withPackerNode` which sits outside the top-level `withEnv`.

**Keyring moved into the role** - `cleanup()` does `rm -rf /tmp/*` at the end of `ubuntu-provision.sh`, so `/tmp/gpg-keys` is gone before any later provisioner. `install_jq` ran before the cleanup so it never noticed, but docker, hashicorp, datadog and nodejs all read keyrings from there and will hit this when they move. Shipping the key in the role's `files/` avoids `/tmp` entirely.

No `ansible_env_vars` are set. An earlier revision carried `ANSIBLE_REMOTE_TEMP` and `ANSIBLE_PIPELINING`; both are gone, see the review discussion below.

Green on all four PR matrix cells, so `amazon-ebs` and `azure-arm` are covered and `become: true` is exercised over real SSH as a non-root user, not just as a no-op in a root container. Play recap `ok=6 changed=4 failed=0`, both goss suites clean, `jq-1.8.2` in the image. A corrupted keyring fails the build, hence no `ignore_errors` on the gpgv task.

One follow-up left out on purpose: there is no `updatecli.d/packer-plugins/ansible.yaml`, so the plugin version pinned in `main.pkr.hcl` will not get bumped for now


provisioning/ansible/roles/install_jq/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
+---
+# Replaces the install_jq() function of provisioning/ubuntu-provision.sh.
+- name: Create a temporary download directory
+  ansible.builtin.tempfile:
+    state: directory
+  register: jq_temp_download_dir
+
+- name: Copy the jq release keyring
+  # Shipped with the role: ubuntu-provision.sh's cleanup() wipes /tmp
```
jayfranco999:
```suggestion
```

provisioning/ansible/roles/install_jq/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+  ansible.builtin.copy:
+    src: jq-keyring.kbx
+    dest: "{{ jq_temp_download_dir.path }}/jq-keyring.kbx"
+    mode: "0644"
+
+- name: Download jq and its detached signature
+  ansible.builtin.get_url:
+    # The signature is published in the jq repository's sig/ tree, not as a release asset
```
jayfranco999:
```suggestion
```

provisioning/ansible/roles/install_jq/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    mode: "0644"
+  loop:
+    - url: "https://github.com/jqlang/jq/releases/download/jq-{{ jq_version }}/{{ jq_binary_file }}"
+      dest: "{{ jq_binary_file }}"
+    - url: "https://raw.githubusercontent.com/jqlang/jq/master/sig/v{{ jq_version }}/{{ jq_binary_file }}.asc"
+      dest: "{{ jq_binary_file }}.asc"
+
+- name: Verify the jq signature
+  ansible.builtin.command:
+    # gpgv looks up a relative --keyring under ~/.gnupg, so pass the full path
```
jayfranco999:
```suggestion
```

build-jenkins-agent-ubuntu.pkr.hcl
```diff
@@ -46,6 +46,23 @@ build {
...
   }
 
+  provisioner "ansible" {
+    playbook_file = "./provisioning/ansible/provision.yml"
+    extra_arguments = [
+      "--extra-vars", "@${var.provision_env_file}",
+      "--extra-vars", "architecture=${var.architecture}",
+    ]
+    ansible_env_vars = [
+      # The docker builder has no IP: the plugin proxies through a local SSH adapter,
```
jayfranco999:
```suggestion
```

build-jenkins-agent-ubuntu.pkr.hcl
```diff
@@ -46,6 +46,23 @@ build {
...
+      # and pipelining avoids the sftp/scp fallback warning on it
```
jayfranco999:
```suggestion
```

build-jenkins-agent-ubuntu.pkr.hcl
```diff
@@ -46,6 +46,20 @@ build {
...
+    playbook_file = "./provisioning/ansible/provision.yml"
+    extra_arguments = [
+      "--extra-vars", "@${var.provision_env_file}",
+      "--extra-vars", "architecture=${var.architecture}",
+    ]
+    ansible_env_vars = [
+      # The play runs as root but `~` resolves to /home/jenkins here, so the default
+      # remote_tmp lands there root owned and breaks the goss `ansible --version`
+      # check, which runs as jenkins. Pin it where a root run would put it anyway.
+      "ANSIBLE_REMOTE_TEMP=/root/.ansible/tmp",
```
dduportal, line 59: Looks like https://oneuptime.com/blog/post/2026-02-21-configure-ansible-remote-temp-directory/view#scenario-4-running-as-a-different-user-with-become.

I would vote for cleaning up the content of "${ANSIBLE_REMOTE_TEMP}" as the last step of this provisioner (or with a shell provisioner right after) instead of specifying a fixed path.

- Using `~/` depends on the `become` and `become_user` directives inside the playbook which are not known here
- Using an absolute path means it will break (or create security holes by allowing unwanted permissions) if we change the user running ansible tasks"

build-jenkins-agent-ubuntu.pkr.hcl
```diff
@@ -46,6 +46,24 @@ build {
...
+    playbook_file = "./provisioning/ansible/provision.yml"
+    extra_arguments = [
+      "--extra-vars", "@${var.provision_env_file}",
+      "--extra-vars", "architecture=${var.architecture}",
+    ]
+    # Rooted in /tmp so it does not depend on which user Ansible connects as, then
+    # removed by the next provisioner so nothing is left in the image.
+    ansible_env_vars = [
+      "ANSIBLE_REMOTE_TEMP=${local.ansible_remote_temp}",
+    ]
```
dduportal, line 59:
```suggestion
```

What is the issue when using the default behavior (and cleaning up after with a shell command or as part of the playbook)?

jayfranco999:
Putting this back in draft while I dig into the user question.

Recap of where I've been going wrong, since it took me a few tries. I kept treating this as "pick a better path for `ANSIBLE_REMOTE_TEMP`" and tried three of them: `/tmp`, then `/root/.ansible/tmp`, then back to the default with a cleanup provisioner. Each one fixed the builder I was thinking about and broke a different one. `/root` died on azure because the connection isn't root there, and the plain default dies on both cloud builders with `unreachable=1` before any task runs.

the point about `ansible_user` is the bit I'd missed. It explains why the path was `/home/jenkins` on all three builders even though the SSH user is `ubuntu` on ebs and something generated on azure. It isn't coming from the communicator at all, it's the provisioner defaulting `user` to whoever runs packer, which on our agents is `jenkins`. And `jenkins` is deliberately not a sudoer, `ubuntu-provision.sh` says so where it creates the account, so `become` can't work either. That fits the "you may have been able to authenticate and did not have permissions" wording in the error better than anything I'd come up with.

So pushed a temporary debugging commit rather than another guess:

- `-vvv` on the provisioner, to see the resolved `ansible_user` and the generated inventory
- a shell provisioner before it, with no `execute_command` override so it runs as the communicator user, dumping `id`, `HOME`, whether `sudo -n` works, and `ls -l` on `/home/jenkins`

That should tell us the actual connecting user per builder instead of me inferring it. Once I have the three values I'll set `user` explicitly, which likely needs conditionals since docker has no `ssh_username` to reuse.

Will report the log output here.

jayfranco999:
`-vvv` and the shell provisioner answered it. On azure:

```
uid=1000(packer) gid=1000(packer) groups=...,27(sudo)
HOME=/home/packer
passwordless sudo: yes
drwxr-x--- 8 jenkins jenkins /home/jenkins
```

then two lines later:

```
ESTABLISH SSH CONNECTION FOR USER: jenkins
-o 'User="jenkins"'
```

So the communicator connects as `packer` and ansible connects as `jenkins`. it's interpolated from the agent running packer.

The two commands side by side show why it dies:

```
echo ~jenkins                            -> (0, '/home/jenkins')
mkdir -p ` echo /home/jenkins/.ansible ` -> (1, 'Permission denied')
```

`~jenkins` is just a passwd lookup so it resolves fine, but the `mkdir` goes through the proxy adapter and runs as the communicator user, which is `packer` here and `ubuntu` on ebs. Neither can write to `/home/jenkins` at 0750. Docker only passed because the adapter execs as root there.

So it was never the temp path, ansible was computing paths for a user it wasn't running as. Pushed a map keyed on `var.image_type`: `root` for docker, `ubuntu` for ebs, `packer` for azure. `ANSIBLE_REMOTE_TEMP` is gone since `~` now resolves to a home the user owns. Debug commit removed.

locals.pkr.hcl
```diff
@@ -48,6 +48,15 @@ locals {
...
     replace(jdk_version, "jdk", "")
   ])
+  # The ansible provisioner defaults its user to the one running packer, not the one
+  # the communicator connects with, so paths like the remote temporary directory get
+  # computed for a user the commands do not actually run as. Set it per builder.
+  ansible_user = {
+    "docker"     = "root"
+    "amazon-ebs" = "ubuntu"
+    "azure-arm"  = "packer"
+  }
```
dduportal: The map "could" be a solution. Syntaxically it works and is simple. 
But it repeat the same information as the provider setups:
- For Ubuntu AWS EBS: https://github.com/jenkins-infra/packer-images/blob/26d2c483880b7f25becc6cfb89bcc104c799c4d5/build-jenkins-agent-ubuntu.pkr.hcl#L8
- For Ubuntu Azure, default is used (see https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm#defaults)
- For Docker, it is the `exec_user` directive which also is set to its default value (ref. https://developer.hashicorp.com/packer/integrations/hashicorp/docker/latest/components/builder/docker#exec_user).

I suggest the making sure that:
- The values from your variable are explicitly specified to both the providers and provisioner so we only have to change them on a single explicit location (to avoid technical debt)
- Use a `local` instead of a `variable`. We don't want the user to dynamically change this. But we still want a local.
- Adapt the name: using it for both provider communicators and ansible provisioner, might be worth setting it up as `ssh_username` for instance WDYT?

jayfranco999: Reworked as suggested. There is now one `local` feeding both the providers and the provisioner:

```hcl
ubuntu_ssh_user = {
  "azure-arm"  = "packer"
  "docker"     = "root"   # The docker builder execs in the container instead of using SSH
  "amazon-ebs" = "ubuntu" # Canonical's AMIs only allow this account to log in
}
```

Used in four places: `ssh_username` on ebs and azure, `exec_user` on docker, and `user` on the ansible provisioner. So the hardcoded `"ubuntu"` is gone and the two values that were previously implicit defaults are now written down.

It was already a `local` rather than a `variable`, so nothing changed there.

On the name, I went with `ubuntu_ssh_user` to mirror the existing `windows_winrm_user` right above it, rather than `ssh_username`, since a local of that name shadows the provider argument. Happy to rename if you prefer yours.

Windows is deliberately untouched. `windows_winrm_user` already does this and converting it is not this PR's job, though I did run `packer validate` on both Windows builders since `locals.pkr.hcl` is shared.

dduportal: LGTM

build-jenkins-agent-ubuntu.pkr.hcl
```diff
@@ -46,6 +48,21 @@ build {
     script           = "./provisioning/ubuntu-provision.sh"
   }
 
+  provisioner "ansible" {
+    playbook_file = "./provisioning/ansible/provision.yml"
+    extra_arguments = [
+      "--extra-vars", "@${var.provision_env_file}",
```
dduportal, line 54: Nit (for subsequent "doc-only" PR): Deserve a comment (with a link to the packer ansible provisionner doc) to explain what is doing the `@`

provisioning/ansible/provision.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Entrypoint for the Ansible provisioning of Linux agents. One role per unit of
+# work, added as functions move out of provisioning/ubuntu-provision.sh.
+- name: Provision Jenkins agent
+  hosts: all
```
dduportal, line 5: Nit (for subsequent "doc-only" PR): Deserve a comment (with a link to the packer ansible provisionner doc) to explain why the playbook applies to `all` hosts from the inventory

build-jenkins-agent-ubuntu.pkr.hcl
```diff
@@ -46,6 +48,21 @@ build {
...
+      "--extra-vars", "architecture=${var.architecture}",
+    ]
+    user = local.ubuntu_ssh_user[var.image_type]
+  }
+
+  # Ansible leaves its scratch directory in the connection user's home
+  provisioner "shell" {
+    execute_command = "{{ .Vars }} sudo -E bash '{{ .Path }}'"
+    inline          = ["rm -rf ~${local.ubuntu_ssh_user[var.image_type]}/.ansible"]
+  }
```
dduportal: Is it still needed now that you solved the Ansible SSH user challenge?

dduportal:
```suggestion
```

main.pkr.hcl
```diff
@@ -17,5 +17,9 @@ packer {
       version = "1.1.4"
       source  = "github.com/hashicorp/docker"
     }
+    ansible = {
+      version = "1.1.4"
```
dduportal, line 21: Nit (for subsequent "doc-only" PR): track with `updatecli`

dduportal (approved):
All good:

- We introduced "explicit" connection users in Packer to ease the Ansible provider setup. Will be usefull to NOT mixed it up with the whole "create `jenkins` user" workflow and avoid permission issues
- The goss test harness passes, showing that this initial implementation works as expected for end users. We'll trigger a packer-image patch release with only this change to be sure we can easily rollback if something unexpected happens
- The last CI check showed that we don't need the custom shell cleanup anymore (as we now have explicit SSH user in which Ansible cleans its temp. directory leaving no state with weird permissions 🥳 )
- 2 things NOT evaluated in this PR/review (food for thought or task for later):
    - Speed improvement. The scope is too tight to see anything valuable as it mostly spends time dowloading (e.g. only depends on network for its order of magnitude)
    - Idempotency/code quality of the `install_jq` role. Will be a task for later when we'll move the role to a reusable collection.
- A few elements for subsequent PR:
  - 2 documentation nit comment, to help reader understand some of the choices (e.g. share the internal knowledge about Ansible provider)
  - 1 updatecli manifest to be created (to track the ansible packer plugin)

PR #3047, opened 2026-09-07
cleanup: remove Windows 2019

lemeurherve:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4954

PR #3058, opened 2026-09-10
cleanup(windows): remove legacy vsconfig

lemeurherve:
This change removes a leftover legacy vsconfig file that was using only in Windows 2019 images.

Amends:
- #3047


lemeurherve:
Unrelated failure, cleanup of a file not used anymore, merging.

